PR #4, opened 2018-11-19
Update RPC protobuf definition

adrianreber:


rst0git:
Should we add `rpc/rpc.proto` in `.gitignore`?

adrianreber:
> Should we add `rpc/rpc.proto` in `.gitignore`?

I updated `.gitignore` to include that temporary file.

rst0git:
LGTM

adrianreber:
Thanks!

Issue #5, opened 2019-01-08
create a release?


vrothberg:
I am currently going through Podman's dependencies (updating the vendor.conf) trying to pin them to releases when possible and available. Some distribution, Debian in particular, are very conservative regarding go packaging as many libraries don't have any release (or no semantic versioning).

Would you consider creating releases of go-criu?

rst0git:
cc: @adrianreber

adrianreber:
> Would you consider creating releases of go-criu?

Sure, no problem.

@rst0git, my first idea was to just use the same version number as CRIU. What do you think?

rst0git:
> @rst0git, my first idea was to just use the same version number as CRIU. What do you think?

Make sense (and it seems like a good practice), especially since the bindings will correspond to a specific CRIU version.

adrianreber:
@vrothberg I created a v3.11 tag and github created a release for us. Does that help you?

@rst0git we should probably document that we will be following CRIU version numbers.

vrothberg:
> @vrothberg I created a v3.11 tag and github created a release for us. Does that help you?

That's awesome. Thanks a lot for the quick response!

> @rst0git we should probably document that we will be following CRIU version numbers.

I think this would help packagers a lot.

PR #6, opened 2019-01-08
Add information about go-criu's version and release scheme

adrianreber:
Signed-off-by: Adrian Reber <areber@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,14 @@ or to just check if at least a certain CRIU version is installed:
 	result, err := c.IsCriuAtLeast(31100)
 ```
 
+## Releases
+
+go-criu will carry the same version number as CRIU. This implies that each
+go-criu release will pull in the necessary changes from CRIU before making a
+release.
+
+The first go-criu release was 3.11 based on CRIU 3.11.
+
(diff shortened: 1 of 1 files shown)


adrianreber:
Related to issue #5

README.md
```diff
@@ -21,6 +21,13 @@ or to just check if at least a certain CRIU version is installed:
 	result, err := c.IsCriuAtLeast(31100)
 ```
 
+## Releases
+
+go-criu will carry the same version number as
+[CRIU](https://github.com/checkpoint-restore/criu). Having the same version
+numbers as CRIU also means that each go-criu release will pull in the
+necessary changes from CRIU before making a release.
+
```
rst0git: Although the protocol buffers guarantees backwards compatibility, should we mention that the go-criu is for version 3.11 onwards?

Could we avoid the repetition of _"the same version numbers as CRIU"_? For example with:
`s/Having the same version numbers as CRIU also means that/This implies that/g`

adrianreber: Changed and force pushed.

rst0git:
LGTM

PR #27, opened 2020-07-17
Support a CRIU binary outside of the default PATH

adrianreber:
Using runc with '--criu /path/to/criu' fails with

msg="CRIU version check failed: exec: \"criu\": executable file not found in $PATH"

This adds SetCriuPath() to go-criu to allow runc to pass the location of the CRIU binary down to go-criu.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: go
-sudo: required
+dist: bionic
 os:
   - linux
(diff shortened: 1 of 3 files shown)


adrianreber:
@kolyshkin I am planing to update runc to use this change. Does the API make sense to you?

PR #28, opened 2020-08-25
Go port of CRIT - Encoding/Decoding 

Pac23:
[7c83bf1](https://github.com/checkpoint-restore/go-criu/pull/28/commits/7c83bf1bde0f90dfb90c20f4c5a5463b35f259fc) once merge wont cause a problem. 
This pr is only for the main encode/decode functionality file and the protobindings.

Travis is not written yet thus passed.


avagin:
I think we can use bazel to generate pb.go  files.

avagin:
or we can write a makefile

crit-go/gocrit/cmdhandler.go
```diff
@@ -0,0 +1,115 @@
...
+import (
+	"crit-go/images"
+	"encoding/json"
+	"fmt"
+	"os"
+)
+
+func Check(e error) {
+	if e != nil {
+		fmt.Println(e)
```
avagin, line 12: all errors have to be printed on stderr

crit-go/gocrit/cmdhandler.go
```diff
@@ -0,0 +1,115 @@
...
+	case "mss":
+		explore_mems(dir)
+	case "rss":
+		explore_rss(dir)
+	}
+}
+
+func Show(inloc string) {
+	fmt.Println("test placeholder - show called")
+
```
avagin, line 65: need to remove this empty line

crit-go/gocrit/cmdhandler.go
```diff
@@ -0,0 +1,115 @@
...
+	case "fds":
+		explore_fds(dir)
+	case "mss":
+		explore_mems(dir)
+	case "rss":
+		explore_rss(dir)
+	}
+}
+
+func Show(inloc string) {
```
avagin, line 63: what does this function do?

crit-go/gocrit/cmdhandler.go
```diff
@@ -0,0 +1,115 @@
...
+		return a pointer to the file or stdin
+	*/
+
+	if inloc == "" {
+		imgfile := os.Stdin
+		return imgfile
+	} else {
+		imgfile, err := os.Open(inloc)
+		if err != nil {
+			fmt.Println("Failed to open input file: %s", err)
```
avagin, line 79: all errors have to be printed on stderr, but I think you need to return error from this function so that the caller can handle it.

crit-go/images/images.go
```diff
@@ -0,0 +1,240 @@
...
+			binary.LittleEndian.PutUint32(bs, uint32(bnvmap))
+			_, err = ouf.Write(bs)
+			checkfile(err, ouf)
+		}
+	}
+	bs := make([]byte, 4)
+	binary.LittleEndian.PutUint32(bs, uint32(magic_val))
+	_, err = ouf.Write(bs)
+	if err != nil {
+		fmt.Println(err)
```
avagin, line 143: it has to exit with a non-zero code if it failed to write into the file.

crit-go/gocrit/cmdhandler.go
Pac23: this is the show function basically decodes with nopl option. similar to the python version.

Pac23:
> I think we can use bazel or make to generate pb.go files.

Make was to be implemented but There is a bug on certain systems.The protoc compiler can cause errors in generating the opts.pb.go file. https://github.com/golang/protobuf/issues/222

crit-go/images/images.go
Pac23: In some places i had to skip the exit. Example : [L233](https://github.com/Pac23/go-criu/blob/59082e70b35a0b2d295de9a7435fbc735d0a4d31/crit-go/images/handle.go#L233) jsonpb complains about the extra data not being a field in the protobinding but will output the .img file. So Exit had to be avoided in all the dump functions of the extra handlers.

crit-go/gocrit/cmdhandler.go
Pac23: Resolved. :+1:

crit-go/gocrit/cmdhandler.go
Pac23: Resolved

crit-go/gocrit/cmdhandler.go
Pac23: Resolved.

crit-go/gocrit/cmdhandler.go
```diff
@@ -0,0 +1,122 @@
...
+	if e != nil {
+		fmt.Println(e)
+		os.Exit(1)
+	}
+}
+
+func checkfile(e error, f *os.File) {
+	if e != nil {
+		f.Close()
+		fmt.Println(e)
```
avagin, line 20: All errors have to be printed on stderr

crit-go/gocrit/cmdhandler.go
avagin: I don't understand how you resolved it. The code is the same.

avagin:
> Make was to be implemented but There is a bug on certain systems.The protoc compiler can cause errors in generating the opts.pb.go file. [golang/protobuf#222](https://github.com/golang/protobuf/issues/222) this thing still persists on some systems especially with the proto2 syntax for generation. [more people experiencing this](https://github.com/golang/protobuf/issues/222#issuecomment-444170144). So until this is resolved upstream i think we should stick to keeping proto bindings on the repo than generating them. Even on my system(apt based) the older protoc compiler did not cause this but the newer one does.

Could you move pg.go files in a separate pr and write a document about how to update them?

crit-go/gocrit/cmdhandler.go
Pac23: just missed this file,other files were fixed.

crit-go/gocrit/cmdhandler.go
Pac23: Resolved.

Pac23:
> > Make was to be implemented but There is a bug on certain systems.The protoc compiler can cause errors in generating the opts.pb.go file. [golang/protobuf#222](https://github.com/golang/protobuf/issues/222) this thing still persists on some systems especially with the proto2 syntax for generation. [more people experiencing this](https://github.com/golang/protobuf/issues/222#issuecomment-444170144). So until this is resolved upstream i think we should stick to keeping proto bindings on the repo than generating them. Even on my system(apt based) the older protoc compiler did not cause this but the newer one does.
> 
> Could you move pg.go files in a separate pr and write a document about how to update them?

Sure, what do you mean by update them though ? as in when the generation is fixed upstream(then how one can update them ?) or something else ?

avagin:
> > I think we can use bazel or make to generate pb.go files.
> 
> Make was to be implemented but There is a bug on certain systems.The protoc compiler can cause errors in generating the opts.pb.go file. [golang/protobuf#222](https://github.com/golang/protobuf/issues/222) this thing still persists on some systems especially with the proto2 syntax for generation. [more people experiencing this](https://github.com/golang/protobuf/issues/222#issuecomment-444170144). So until this is resolved upstream i think we should stick to keeping proto bindings on the repo than generating them. Even on my system(apt based) the older protoc compiler did not cause this but the newer one does.

I don't understand what "this thing still persists on some systems". What system are you talking about here? You gave two links on github issues, but both these issues are closed and both were updated about two year ago. This means nobody experience them any more, doesn't it?

avagin:
> Sure, what do you mean by update them though?

I mean we can add a new image in CRIU or we can add a new field in an existing image and then we will need to update pb.go files in this project and we need an instruction how to do this. In ideal case, we need a script while will do this.

crit-go/gocrit/cmdhandler.go
```diff
@@ -0,0 +1,120 @@
...
+	if e != nil {
+		fmt.Fprintln(os.Stderr, e)
+		os.Exit(1)
+	}
+}
+
+func checkfile(e error, f *os.File) {
+	if e != nil {
+		f.Close()
+		fmt.Fprintln(os.Stderr, e)
```
avagin, line 20: This error message will be non-informative. It will not say what operations triggers an error. I don't like these checkfile and Check functions, this isn't go-style.

crit-go/cmd/root.go
```diff
@@ -0,0 +1,38 @@
...
+// rootCmd represents the base command when called without any subcommands
+
+var rootCmd = &cobra.Command{
+	Use:   "crit-go",
+	Short: "CRIT is a feature-rich replacement for existing criu show",
+	Long: `CRIT is a feature-rich replacement for existing "criu show". 
+	This version is written in Go for usage with Go codebase, crit is also available in python
+	for more information visit https://criu.org/CRIT `,
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Println("crit executed")
```
avagin, line 21: all errors and debug messages have to be printed on stderr. I would recomend to use one of logging libraries instead of using fmt.Println.

crit-go/cmd/root.go
```diff
@@ -0,0 +1,38 @@
...
+	},
+}
+
+func Execute() {
+	if err := rootCmd.Execute(); err != nil {
+		fmt.Println(err)
```
avagin, line 27: stderr

Pac23: what i will do is simply fix the cli on this branch itself will be easy to keep track of things now that the pb files which made up the majority of this branch are removed.

crit-go/cmd/root.go
Pac23: > > Sure, what do you mean by update them though?
> 
> I mean we can add a new image in CRIU or we can add a new field in an existing image and then we will need to update pb.go files in this project and we need an instruction how to do this. In ideal case, we need a script while will do this.

there is already a pr to update the images,which uses the go-out flag so that we can generate these using make on the system itself.

Pac23:
> > > I think we can use bazel or make to generate pb.go files.
> > 
> > 
> > Make was to be implemented but There is a bug on certain systems.The protoc compiler can cause errors in generating the opts.pb.go file. [golang/protobuf#222](https://github.com/golang/protobuf/issues/222) this thing still persists on some systems especially with the proto2 syntax for generation. [more people experiencing this](https://github.com/golang/protobuf/issues/222#issuecomment-444170144). So until this is resolved upstream i think we should stick to keeping proto bindings on the repo than generating them. Even on my system(apt based) the older protoc compiler did not cause this but the newer one does.
> 
> I don't understand what "this thing still persists on some systems". What system are you talking about here? You gave two links on github issues, but both these issues are closed and both were updated about two year ago. This means nobody experience them any more, doesn't it?

i wouldn't conclude no body experiences them.Lot of things in general with protobuf are broken and if someone has a slightly misconfigured system this error imo might reappear or if someone used a different version of protoc compiler in one of the links if you look someone says it appears on centos but not on macos.I do get that this was two years ago but  I personally experienced this error on my  Ubuntu 18.04LTS system. I know the fix for this is using the `m` flag in `go out` but there are a lot of unknown variables here that might cause a wrongly generated `pb.go.` file on someones system.

Pac23:
> > Sure, what do you mean by update them though?
> 
> I mean we can add a new image in CRIU or we can add a new field in an existing image and then we will need to update pb.go files in this project and we need an instruction how to do this. In ideal case, we need a script while will do this.

i am still bit confused as to what you mean. Anyways there is already a pr to update the protobinding images in the main repo the pr adds a `option go_package = "/protobindings";` flag to the current proto images. Is this what you mean by add a new field in a existing image ?

crit-go/cmd/root.go
Pac23: is there a recommended logging library for that criu uses or can i use any logging lib ?

crit-go/gocrit/cmdhandler.go
Pac23: okay, i will remove this.

crit-go/images/images.go
```diff
@@ -0,0 +1,240 @@
...
+	/*
+	   Convert criu image from binary format to dict(json).
+	   Takes a file-like object to read criu image from.
+	   Returns criu image in dict(json) format.
+	*/
+	magicjson()
+	m := magichandler(f)
+	image["magic"] = m
+	switch {
+	case m == "GHOST_FILE":
```
avagin, line 47: we can have a map of handlers and the code will looks like this:
handler, ok := handlers[m];
if !ok {
  		handler := &entry_handler{}
}
handler.m = m
image["entries"] = handler.Load(f, pretty, nopl)

crit-go/gocrit/cmdhandler.go
avagin: you need to resolve it here. Each patch have to be clean to be merge.

crit-go/images/images.go
Pac23: this will require type assertion for the struct methods like load etc to work which basically circles back to a type switch.

crit-go/cmd/root.go
Pac23: Resolved with logging

crit-go/cmd/root.go
Pac23: refactored to use logging

crit-go/gocrit/cmdhandler.go
Pac23: i have removed these functions replacing them with logging

adrianreber:
I do not see any Makefile changes? Is this being run through the linter and build during CI runs?

PR #29, opened 2020-08-26
changed magic.json to turn byname key:vals into string.

Pac23:
This changes certain current magic-gen.json fields to string.This is a prerequisite for the go port of crit.Minor Change from the original file.


adrianreber:
This breaks CI https://travis-ci.org/github/checkpoint-restore/go-criu/jobs/727697039

When you submitted it was still working, now it is broken. Maybe you could have a look at it.

PR #32, opened 2020-10-11
go.{mod,sum}: update

kolyshkin:
Commit ba3a445dda6941ba82 added use of github.com/thoas/go-funk but it
was not reflected in go.mod.

This commit is brought to you by

	go mod tidy

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,4 +2,7 @@ module github.com/checkpoint-restore/go-criu/v4
 
 go 1.13
 
-require github.com/golang/protobuf v1.3.5
+require (
+	github.com/golang/protobuf v1.3.5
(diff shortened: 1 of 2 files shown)


rst0git (approved):

adrianreber:
@kolyshkin Thanks. I did not update `go.mod` because I thought that `go-funk` is only used in `magic-gen` which is not really used, yet. But I guess I misunderstood Go modules once again.

kolyshkin:
> I thought that go-funk is only used in magic-gen which is not really used, yet. But I guess I misunderstood Go modules once again.

No one understands Go modules in their entirety :trollface: 

Closing this one in favor of https://github.com/checkpoint-restore/go-criu/pull/37

PR #35, opened 2020-11-22
Detect if process is closed during migration in phaul using pidfd

ZeyadYasser:
This allows detection if a process was closed during migration reliably using ```pidfds``` if ```pidfd_open``` syscall is supported.

CRIU already detects PID reuse using task's creation time internally, this further enforces the detection through P.Haul during migration using ```pidfds```.

One short coming of this approach is that there might be a race condition between the call to ```isPidClosed()``` and ```criu.PreDump()``` which would happen if the process was closed exactly after the check and right before the pre-dump, if a process managed to reuse that pid in this duration it should be detected in CRIU using task's creation time, it is very rare but might happen.

Related issue: https://github.com/checkpoint-restore/criu/issues/717

### DIFF
--- a/phaul/pidfd.go
+++ b/phaul/pidfd.go
@@ -1,0 +1,41 @@
+package phaul
+
+import (
+	"fmt"
+	"golang.org/x/sys/unix"
+)
+
+func pidfdOpen(pid int) (int32, error) {
+	r0, _, e1 := unix.Syscall(unix.SYS_PIDFD_OPEN, uintptr(pid), uintptr(0), uintptr(0))
+	pidfd := int32(r0)
+
+	if e1 != 0 {
+		return -1, e1
+	}
+
+	return pidfd, nil
+}
+
+// isPidClosed function
+// Checks that the process with the given pidfd is not closed.
+// When the process that the pidfd refers to terminates, poll
+// indicates the file descriptor as readable.
+func isPidClosed(pidfd int32) bool {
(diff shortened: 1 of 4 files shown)


phaul/pidfd.go
```diff
@@ -0,0 +1,44 @@
...
+	return pidfd, nil
+}
+
+// isPidClosed function
+// Checks that the process with the given pidfd is not closed.
+// When the process that the pidfd refers to terminates, poll
+// indicates the file descriptor as readable.
+func isPidClosed(pidfd int32) bool {
+	pollfds := []unix.PollFd{
+		unix.PollFd{
```
kolyshkin, line 25: AFAICS this line is redundant, i.e. we can do something like

```go
pollfds := []unix.PollFd{
   { Fd: pidfd, Events: unix.POLLIN },
}
```

ZeyadYasser: Updated

Snorch:
1) You only check that root process in the dumped process tree is still the same. But that's not enough.
    - You should check it for each process of the dumped process tree.
2) You should not just fail if you've detected pid-reuse.
    - You should make criu know about it and make it skip memory images of this pid from previous iterations.
3) You should not just skip non-working pid reuse like "(skipping pidfd based pid reuse detection)"
    - If pid-reuse does not work there is a risk of memory corruption or exploit activity due to migration, we should just fail in this case.

PR #37, opened 2020-12-03
Fix go.mod

kolyshkin:
Add a separate go.mod to crit-go/magic-gen since it's a separate
binary which is not part of the library/package.

Generated by:
```
	cd crit-go/magic-gen
	touch go.mod
	go mod tidy
```


adrianreber:
Perfect. Thanks.

PR #38, opened 2020-12-04
go.mod: bump github.com/golang/protobuf to v1.4.3

kolyshkin:
_(this is currently based on #37)_

This should fix the following deprecation warning (seen in runc):
    
> SA1019: package github.com/golang/protobuf/proto is deprecated: Use
> the "google.golang.org/protobuf/proto" package instead.  (staticcheck)
    
This commit brought to you by

```    
            go get github.com/golang/protobuf@v1.4
            go mod tidy
```


kolyshkin:
@adrianreber I think this results in some backward incompatibility, and thus requires a (minor) release (`v4.2.0` I guess?)

kolyshkin:
Hmm maybe we should switch to using `google.golang.org/protobuf/proto` instead. Let me take a look...

kolyshkin:
> Hmm maybe we should switch to using google.golang.org/protobuf/proto instead. Let me take a look...

Or not yet, since distros (e.g. fedora) are still using the old package ...

kolyshkin:
OK, I think this PR is good to go. https://github.com/golang/protobuf README says

> Versions v1.4 and later of github.com/golang/protobuf are implemented in terms of google.golang.org/protobuf. Programs which use both modules must use at least version v1.4 of this one.

adrianreber:
As you have the most experience with Go modules from anyone involved in go-criu I am just trusting you that this is correct. Thanks for helping us out.

PR #39, opened 2020-12-20
CI: Migrating to GitHub Actions

rst0git:
After December 31st, 2020 travis-ci.org will be switched to a read-only platform. This PR converts the integration tests from Travis CI to GitHub Actions.


adrianreber:
Do you know why your configured action is not running? What about adding golang 1.15 also as a test target.

adrianreber:
Thanks. Let's merge this to see the results.

PR #46, opened 2021-02-10
gha: add golangci-lint job, fix warnings; bump API version

kolyshkin:
This
 - reworks the source layout to ensure linters don't choke;
 - slightly improves Makefile and running tests;
 - introduces golangci-lint from Makefile as well as for GitHub Actions;
 - fixes all golangci-lint warnings found;
   - :warning: this includes bumping API to v5 since we changed the prototypes;
 - adds a small `go mod tidy` verify job to GHA;
 - enables some more linters and fixes their warnings (suggested by @rst0git)

See individual commits for more details.

Description for changelog:
```
 - API bumped to v5
 - methods (such as Dump, Restore etc.) now require a pointer to rpc.CriuOpts as the first argument
```


kolyshkin:
OK, this can't be done until we have a mix of .c and .go files. Trying to fix...

kolyshkin:
@adrianreber @rst0git PTAL. Unfortunately I had to change the public API and thus bump v4 to v5 -- I see no way around it.

adrianreber:
Thanks. Looks good to me. I have no problem doing a v5 based release. @rst0git what do you think about these changes?

Makefile
```diff
@@ -1,55 +1,46 @@
...
-ifeq ($(GOBIN),)
-	GOBIN := $(FIRST_GOPATH)/bin
-endif
 
-all: build test phaul phaul-test
+all: build test phaul-test
 
 lint:
-	@golint -set_exit_status . test phaul
+	golangci-lint run ./...
```
rst0git, line 7: What about enabling some of the linters presets?

```
golangci-lint run -p bugs,performance,unused,format ./...
```

kolyshkin: I think this should be addressed separately (and it's better/easier to do so via `.golangci.yml`), once this is merged.

kolyshkin: @rst0git implemented; PTAL

kolyshkin:
(CI for 1.16-rc1 is broken as it was removed; fix in #47)

Makefile
kolyshkin: This probably makes the patchset slightly harder to review -- let me know if you want a separate PR.

rst0git (approved):

Makefile
rst0git: Thanks!

adrianreber:
@rst0git Do you want to tag the new release? 5.0.0?

PR #43, opened 2021-02-02
Update protobuf files and go mod/sum

kolyshkin:
This is a followup to https://github.com/checkpoint-restore/go-criu/pull/38, addressing https://github.com/checkpoint-restore/go-criu/issues/36.

This updates `stats/stats.pb.go` and `rpc/rpc.pb.go` to versions generated by latest protoc-gen-go. This makes them compatible with github.com/golang/protobuf@v1.4 as well as google.golang.org/protobuf/proto.

This also updates go.mod and go.sum, adding golang.org/x/sys/unix (used by phaul) and google.golang.org/protobuf (used by *.pb.go).

See more details in commit messages.

PR #49, opened 2021-02-17
README: prepare for v5.0.0

kolyshkin:
In anticipation of v5.0.0 release, add some info to README (and fix some nits while at it).

 - README: no links in header, highlight go code
 - README: provide a complete example that includes import statement
 - README: add info about v5, use a table
 - README: add link to go doc

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,12 @@ As go-criu is imported in other projects and as Go modules are expected
 to follow Semantic Versioning go-criu will also follow Semantic Versioning
 starting with the 4.0.0 release.
 
-4.0.0 is based on CRIU 3.14
+The following table shows the relation between go-criu and criu versions:
+
+| Major version  | Latest release | CRIU version |
+| -------------- | -------------- | ------------ |
+| v5             | 5.0.0          | 3.15         |
+| v4             | 4.1.0          | 3.14         |
 
 ## How to contribute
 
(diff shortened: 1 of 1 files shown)


kolyshkin:
Note that I am testing v5 at https://github.com/opencontainers/runc/pull/2807 (using a `replace` trick in go.mod and my own fork) and it appears to work fine.

Funny thing it, I just discover no one is using methods like `Dump()`, `Restore()` etc.

adrianreber:
> Note that I am testing v5 at [opencontainers/runc#2807](https://github.com/opencontainers/runc/pull/2807) (using a `replace` trick in go.mod and my own fork) and it appears to work fine.

Nice.

> Funny thing it, I just discover no one is using methods like `Dump()`, `Restore()` etc.

The main reason for that is, I guess, that runc is basically the only user doing the actual dump and restore (in Go). There was just no need/motivation to pull it out from runc into go-criu. I actually looked at it, but looking was all I did so far. Handling the callbacks (https://github.com/opencontainers/runc/blob/master/libcontainer/container_linux.go#L1665) is one the things which require to have a well designed API in go-criu. I think this currently does not exist.

The version check is used by other projects (Podman, CRI-O PR). So it made sense to move that to go-criu.

kolyshkin:
To my comment about no users of Dump(), Restore() etc -- if there are no users, we can keep v4 and not switch to v5 as we don't break anything/anyone by changing the API. Let me know what you think @adrianreber 

Yeah, maybe we can work on moving some CRIU stuff from runc to here and designing a decent API -- but only after runc 1.0 final is released (which I guess is a few months away).

Another idea (which is still pretty blurry in my head) is to write a small CLI tool which uses this library and replaces criu -- not that criu CLI needs replacement, but to have another user of this code (may be also helpful in writing tests etc)

kolyshkin:
@adrianreber can we please have a release, so that runc uses a tagged version of the library (https://github.com/opencontainers/runc/pull/2807 is currently blocked because there is no tagged version that incorporates #43)

adrianreber (approved):

adrianreber:
@kolyshkin I approved it, but it does not automatically merge it. It still says that all commits have to be signed.

adrianreber (approved):

PR #53, opened 2021-06-29
Add 'vendor' target to Makefile and run it

adrianreber:


adrianreber:
@kolyshkin Does adding 'make vendor' make sense or is it unnecessary?

rst0git:
I think it makes sense. This is similar to `make vendor` in [podman](https://github.com/containers/podman/blob/599b7d746a6ccad6ebbd70abb72be30bead094a6/Makefile#L272).

Makefile
```diff
@@ -54,4 +54,9 @@ rpc/rpc.pb.go: rpc/rpc.proto
 stats/stats.pb.go: stats/stats.proto
 	protoc --go_out=. $^
 
-.PHONY: build test phaul-test test-bin clean lint
+vendor:
+	GO111MODULE=on go mod tidy
+	GO111MODULE=on go mod vendor
+	GO111MODULE=on go mod verify
```
rst0git, line 60: Perhaps we should use `$(GO)` instead of `go`?

adrianreber: done

rst0git (approved):
LGTM

PR #54, opened 2021-07-02
Update to latest CRIU RPC protobuf definition

adrianreber:
This updates to the latest CRIU RPC protobuf definition from criu-dev.

Now, that CRIU protobuf files are licensed as 'MIT', the protobuf definition can also be included in the repository.

Main motivation for this change is to provide LsmMountContext in runc.


rst0git:
> This updates to the latest CRIU RPC protobuf definition from criu-dev.

Should we wait for the 3.16 release before merging it?

adrianreber:
> > This updates to the latest CRIU RPC protobuf definition from criu-dev.
> 
> Should we wait for the 3.16 release before merging it?

I would prefer not to wait. I do not expect any major changes in CRIU concerning RPC and once this is merged I would make a 5.1.0 release which I need for a runc pull request. If we need any further changes we can always create a new release if necessary after 5.1.0.

rst0git (approved):

PR #55, opened 2021-10-10
Add stats.proto to repo and regenerate with latest protoc-gen-go

zhsj:


adrianreber:
> > What is the motivation for this change?
> 
> Regenerate with latest protoc-gen-go, so get rid of github.com/golang/protobuf in go.mod
> 
> The rpc.pb.go has already not imported github.com/golang/protobuf, but the stats.pb.go still has.

Sounds good. Can you include this information in the corresponding commit message?

zhsj:
> Sounds good. Can you include this information in the corresponding commit message?

Done

PR #57, opened 2021-10-29
Move statistics read funtion to stats/

adrianreber:
This moves the statistics read function to stats/. Adds a function for reading restore statistics and also adds a check for the image file magic.


rst0git (approved):
LGTM

PR #60, opened 2021-12-16
Add feature checking support to go-criu

adrianreber:
This adds the feature checking interface with the same semantics as the feature checking through libcriu.


rst0git (approved):
LGTM

Issue #64, opened 2022-04-14
Rework magic-gen.go into standalone script


snprajwal:
magic-gen need not be a module by itself, it can easily be rewritten into a standalone script that can be run directly without building. The logic can been simplified and logging made more verbose, to enable better debugging. The test in the makefile also uses a `samplemagic` file which need not be retained.

PR #65, opened 2022-04-14
magic-gen: move to standalone script

snprajwal:
Simplified logic to enable better debugging and remove dependencies.
Edited makefile to run tests without leaving any files behind.


snprajwal:
fixes #64

.github/workflows/main.yml
```diff
@@ -33,8 +33,6 @@ jobs:
       run: |
         # Run actual test as root as it uses CRIU.
         sudo make test phaul-test
-        # This builds crit-go
-        sudo make -C crit-go/magic-gen build magicgen test
```
adrianreber, line 36: Please keep the test running and point it to the new location.

.github/workflows/verify.yml
```diff
@@ -12,13 +12,3 @@ jobs:
...
         # Only show new issues for a pull request.
         only-new-issues: true
-
-  gomod:
-    runs-on: ubuntu-20.04
-    steps:
-    - uses: actions/checkout@v2
-    - name: verify go.mod/go.sum
-      run: |
-        go mod tidy
```
adrianreber, line 22: Please do not remove `go mod tidy` for the main directory.

rst0git:
@snprajwal you can use `git mv` to move a file and update the git index.

adrianreber:
You should not have any fixup commits in your pull request. If you fixed one of your previous commits you can directly modify your original commit.

Having a clear description why this change is needed would be good to have in one of the commits. Maybe the information you added to the corresponding issue.

Another thing which came to my mind. Is it actually necessary to have a `json` file. Couldn't you just create some go file which can then directly be used from future Go code with reading it as JSON? Not sure. Just an idea.

snprajwal:
That makes sense, instead of a json, we can generate a map in Go and provide that under the `magic` package itself. Will edit the commits :+1:

snprajwal:
Sorry for multiple pushes, the changes are done. The script is working as expected.

scripts/Makefile
```diff
@@ -0,0 +1,17 @@
...
+magic-gen: clean magic.h
+	$(GO) run magicgen.go magic.h ../magic/magic.go
+	rm -f magic.h
+
+magic.h:
+	curl -s https://raw.githubusercontent.com/checkpoint-restore/criu/criu-dev/criu/include/magic.h -o magic.h
+
+test:
+	@./magicgen-test.sh
+	@rm -f input.h output.go expected.go
```
rst0git, line 12: If the test fails it would be helpful to be able to look at the `input.h`, `output.go`, `expected.go` files. We could add them in `.gitignore` instead of removing them.

scripts/magicgen-test.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
```
rst0git, line 2: Could you please add `set -x`? This would make it easier to debug if the test fails in CI.

scripts/magicgen-test.sh
```diff
@@ -0,0 +1,39 @@
...
+EOF
+
+go run magicgen.go input.h output.go
+cmp output.go expected.go
+if [[ $? -eq 0 ]]
+then
+	echo "PASS"
+	exit 0
+else
+	echo "FAIL"
```
rst0git, line 37: It would be helpful to run `diff expected.go output.go` if the test fails.

scripts/Makefile
```diff
@@ -0,0 +1,17 @@
+GO ?= go
+
+magic-gen: clean magic.h
+	$(GO) run magicgen.go magic.h ../magic/magic.go
+	rm -f magic.h
```
rst0git, line 5: Perhaps we could add `magic.h` in `.gitignore` instead of removing it?

magic/magic.go
```diff
@@ -0,0 +1,158 @@
+// Code generated by magicgen. DO NOT EDIT.
```
rst0git, line 1: If we auto-generate this file, perhaps we don't need to include it in the repository. We could also add it in `.gitignore`.

scripts/Makefile
snprajwal: The `exit 1` in the test script ensures that the files are not deleted if the test fails. They are removed only when the test runs successfully.

magic/magic.go
snprajwal: I don't think we need to generate it every time we build. The only time we'll have to re-generate will be if the original `magic.h` in the criu repo changes. Do we add something to the parent makefile to generate this when `make` is invoked?

rst0git: Auto-generated files are usually excluded from the git version control.

If you think there is a reason to keep track of the changes in this file, we need to fix the following two problems:

1. Every time we run `make -C scripts/` the content of the `magic/magic.go` file changes
2. Running `make clean -C scripts/`  would remove `magic/magic.go`

scripts/Makefile
rst0git: You are right, the files are removed only when the test runs successfully.

It would still be useful to add them in `.gitignore`. Otherwise they appears as untracked files and may be accidentally be added in a git commit.
```
$ git status
...
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	expected.go
	input.h
	output.go
```

magic/magic.go
snprajwal: Makes sense, let's exclude it

rst0git (approved):
LGTM

adrianreber:
@rst0git If you think this is the right thing to do, please merge it. Looks good from a final quick view over it.

PR #66, opened 2022-05-26
[GSoC 2022] Port CRIT functionalities

snprajwal:
# Overview

This PR intends to provide a fully functional Go library that enables users to directly operate on CRIU image files natively within their Go programs. It also provides a CLI app that does the same thing as the Python `crit`.

# Implementation

The library provides a `CriuImage` struct with the entries generalised as a slice of `proto.Message` interfaces. This lets us use the same struct to hold all kinds of images, with minimal type assertion while converting between formats. Once the struct object itself is created, no type assertion is necessary as Go provides the concrete proto type for each member of the slice. The extra data for all entries is stored as a string, as almost all of it is base64 encoded data. Custom JSON marshalers and unmarshalers are used to separate the extra data from the protobuf entries during encoding and decoding.

# Features
- [x] Decoder (load binary into Go struct)
- [x] Encoder (write Go struct into binary file)
- [x] JSON Parser (load JSON file into Go struct)
- [x] Dir explorer (load {ps,fds,mems,rss} data into Go struct)

# CLI
- [x] `crit decode`
- [x] `crit encode`
- [x] `crit show`
- [x] `crit info`
- [x] `crit x ps`
- [x] `crit x fds`
- [x] `crit x mems`
- [x] `crit x rss`

# Additional tasks
- [ ] Unit tests
- [ ] Integration test
- [ ] Inline comments for LSP hover
- [ ] Documentation on CRIU website


.gitignore
```diff
@@ -8,3 +8,5 @@ scripts/*.h
 scripts/expected.go
 scripts/output.go
 magic/magic.go
+crit/images/*.pb.go
+crit/images/*.proto
```
rst0git, line 12: It would be good to add the `*.proto` files to avoid downloading them every time.

crit/Makefile
```diff
@@ -0,0 +1,26 @@
...
+proto_path := ./images
+# Generate string of all .proto filenames
+proto_files := $(sort $(subst $(proto_path)/,,$(wildcard $(proto_path)/*.proto)))
+# Generate M flag to specify import path for all .proto files
+# and replace all spaces with commas to use with go_opt flag
+comma := ,
+proto_opts := $(subst $() $(),$(comma),$(patsubst %,M%=$(import_path),$(proto_files)))
+
+get-proto:
+	git clone https://github.com/checkpoint-restore/criu criu-temp
```
rst0git, line 15: You can add `--depth=1` and `-b master` to download only the latest commit from the master branch.

crit/Makefile
```diff
@@ -0,0 +1,26 @@
...
+	rm -rf criu-temp
+
+gen-proto:
+	@protoc \
+		--proto_path=$(proto_path) \
+		--go_out=$(proto_path) \
+		--go_opt=paths=source_relative,$(proto_opts) \
+		$(proto_files)
+
+.PHONY = gen-proto get-proto
```
rst0git, line 26:
```Makefile
.PHONY: gen-proto get-proto
```

crit/Makefile
```diff
@@ -12,9 +12,10 @@ comma := ,
 proto_opts := $(subst $() $(),$(comma),$(patsubst %,M%=$(import_path),$(proto_files)))
 
 get-proto:
-	git clone https://github.com/checkpoint-restore/criu criu-temp
+	rm ./images/*.proto || true
+	git clone --depth 1 --branch master https://github.com/checkpoint-restore/criu criu-temp
 	cp criu-temp/images/*.proto ./images/
-	rm -rf criu-temp
+	rm -R criu-temp
```
rst0git, line 18: Why do we need this change?

.gitignore
```diff
@@ -9,4 +9,3 @@ scripts/expected.go
 scripts/output.go
 magic/magic.go
 crit/images/*.pb.go
-crit/images/*.proto
```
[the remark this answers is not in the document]
rst0git: No need to squash the code into a single commit. The reason for applying the changes in the previous commit is to make the pull request easier to review. Thus, each logical change is in a [separate commit](https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#separate-your-changes). For example, it would good to add the *.proto files in a separate commit.

crit/images/handler.go
```diff
@@ -0,0 +1,146 @@
...
+	"errors"
+	"fmt"
+
+	"google.golang.org/protobuf/proto"
+)
+
+func ProtoHandler(magic string) (proto.Message, error) {
+	switch magic {
+		case "FS":
+			return &FsEntry{}, nil
```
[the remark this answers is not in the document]
snprajwal: Got it, adding `make -C crit gen-proto` to the `build` target in the root Makefile fixes this.

crit/cmd/cli.go
```diff
@@ -0,0 +1,165 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/checkpoint-restore/go-criu/v5/crit"
+	"github.com/spf13/cobra"
```
rst0git, line 11: Would you be able to run `make vendor` and include the changes with this commit as well?

crit/Makefile
```diff
@@ -0,0 +1,27 @@
...
+import_path := github.com/checkpoint-restore/go-criu/crit/images
+# Path to .proto source files
+proto_path := ./images
+# Generate string of all .proto filenames
+proto_files := $(sort $(subst $(proto_path)/,,$(wildcard $(proto_path)/*.proto)))
+# Generate M flag to specify import path for all .proto files
+# and replace all spaces with commas to use with go_opt flag
+comma := ,
+proto_opts := $(subst $() $(),$(comma),$(patsubst %,M%=$(import_path),$(proto_files)))
+
```
rst0git, line 13: It would be good to add Makefile target to build an executable for `crit`
```Makefile
all: gen-proto bin/crit

bin/crit: cmd/cli.go
	go build -o $@ $<
```

crit/Makefile
```diff
@@ -0,0 +1,27 @@
...
+get-proto:
```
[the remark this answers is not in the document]
snprajwal: > It would be good to commit crit/images/*.pb.go files.

The original `criu` repo maintains only the `.proto` files, and uses the Makefile to generate the bindings for both Python and C. I thought it would be better to stick to the same system here. I have added the target to the parent Makefile [here](https://github.com/checkpoint-restore/go-criu/pull/66/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R17), which also ensures that the tests pass. I'm unsure if there are any other benefits to committing the `.pb.go` files.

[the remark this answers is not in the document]
snprajwal: I've added `crit/images` to `skip-dirs` in `.golangci.yml`. I just noticed that there was a typo that stopped it from working. The YAML section should be `skip-dirs` with a hyphen, not underscore. This should fix the lint errors.

.golangci.yml
```diff
@@ -1,5 +1,5 @@
 run:
-  skip_dirs:
+  skip-dirs:
```
rst0git, line 2: @snprajwal Could open a new pull request with this change?

crit/cmd/cli.go
```diff
@@ -71,11 +72,17 @@ var encodeCmd = &cobra.Command{
 	Short: "Convert JSON to binary file",
 	Long:  "Convert the input JSON to a CRIU image file.",
 	Run: func(cmd *cobra.Command, args []string) {
-		c = crit.New(inputFilePath, outputFilePath, inputDirPath, exploreType, pretty, noPayload)
-		_, err := c.Dump()
+		c = crit.New(inputFilePath, outputFilePath,
+			inputDirPath, exploreType, pretty, noPayload)
+		// Convert JSON to Go struct
+		img, err := c.Parse()
 		if err != nil {
```
rst0git, line 79: This error check could be on a single line:
```go
if img, err := c.Parse(); err != nil {
}
```

crit/cmd/cli.go
```diff
@@ -33,7 +33,8 @@ var decodeCmd = &cobra.Command{
 	Long: `Convert the input binary image to JSON and writes it to a file.
 If no output file is provided, the JSON is printed to stdout.`,
 	Run: func(cmd *cobra.Command, args []string) {
-		c = crit.New(inputFilePath, outputFilePath, inputDirPath, exploreType, pretty, noPayload)
+		c = crit.New(inputFilePath, outputFilePath,
+			inputDirPath, exploreType, pretty, noPayload)
```
rst0git, line 37: Would it be possible to move the changes that split long lines in a separate commit?
It would make it much easier to review.

crit/crit.go
```diff
@@ -50,12 +58,36 @@ func (c *crit) Load() (*CriuImage, error) {
...
+
+	img := CriuImage{}
+	if err = json.Unmarshal(jsonData, &img); err != nil {
+		return nil, errors.New(fmt.Sprint("Error processing JSON: ", err))
+	}
+
+	return &img, nil
+}
+
+func (c *crit) Dump(img *CriuImage) error {
```
rst0git, line 83: The name `Dump` is slightly confusing here because we use this name for the checkpoint functionality in CRIU.
Perhaps `Decode` could be more self-explanatory?

crit/dump-extra.go
```diff
@@ -0,0 +1,116 @@
...
+	"encoding/base64"
+	"encoding/binary"
+	"encoding/json"
+
+	"github.com/checkpoint-restore/go-criu/v5/crit/images"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
+
+func dumpPipesData(extra string) ([]byte, error) {
```
rst0git, line 13: `decodePipesData`?

crit/dump-extra.go
```diff
@@ -0,0 +1,116 @@
...
+	return base64.StdEncoding.DecodeString(extra)
+}
+
+func dumpSkQueues(extra string) ([]byte, error) {
```
rst0git, line 17: `decodeSkQueues`?

crit/dump-extra.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	// Round off to nearest 32-bit multiple
+	roundedSize := len(extraPayload)
+	// Append zeroes for remaining bytes
+	extraPayload = append(extraPayload, make([]byte, roundedSize-len(extraPayload))...)
+
+	return extraPayload, nil
+}
+
+func dumpIpcMsg(extra string) ([]byte, error) {
```
rst0git, line 78: `decodeIpcMsg`?

crit/dump-extra.go
```diff
@@ -0,0 +1,116 @@
...
+	extraSize := len(extraEntries) * 2
+	// Round off to nearest 64-bit multiple
+	roundedSize := (extraSize/8 + 1) * 8
+	// Append zeroes for the remaining bytes
+	extraPayload = append(extraPayload, make([]byte, roundedSize-extraSize)...)
+
+	return extraPayload, nil
+}
+
+func dumpIpcShm(extra string) ([]byte, error) {
```
rst0git, line 65: `decodeIpcShm`?

crit/dump-extra.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	return append(inqBytes, outQBytes...), nil
+}
+
+func dumpBpfmapData(extra string) ([]byte, error) {
+	return base64.StdEncoding.DecodeString(extra)
+}
+
+func dumpIpcSem(extra string) ([]byte, error) {
```
rst0git, line 43: `decodeIpcSem`?

crit/dump-extra.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	outQBytes, err := base64.StdEncoding.DecodeString(extraPayload.OutQ)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(inqBytes, outQBytes...), nil
+}
+
+func dumpBpfmapData(extra string) ([]byte, error) {
```
rst0git, line 39: `decodeBpfmapData`?

crit/dump-extra.go
```diff
@@ -0,0 +1,116 @@
...
+
+func dumpPipesData(extra string) ([]byte, error) {
+	return base64.StdEncoding.DecodeString(extra)
+}
+
+func dumpSkQueues(extra string) ([]byte, error) {
+	return base64.StdEncoding.DecodeString(extra)
+}
+
+func dumpTcpStream(extra string) ([]byte, error) {
```
rst0git, line 21: `decodeTcpStream`?

crit/dump.go
```diff
@@ -0,0 +1,166 @@
...
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/checkpoint-restore/go-criu/v5/magic"
+	"google.golang.org/protobuf/proto"
+)
+
+func dumpImg(img *CriuImage, f *os.File) error {
```
rst0git, line 14: `decodeImg`?

crit/cmd/cli.go
snprajwal: Making it a one-liner reduces the scope of `img` to the `if` block. I'm using it in line 83 for `c.Dump(img)`

crit/crit.go
snprajwal: >The name Dump is slightly confusing here because we use this name for the checkpoint functionality in CRIU.

Technically, CRIU's `dump` and the `Dump()` function here are both generating binary image files. In that sense, I think the naming is consistent with the action being performed. Calling it `Encode()` might confuse those who use the lib as an import, because `crit encode` is a combination of `Parse()` and `Dump()`.

Also, we are explicitly stating `*crit` as a receiver for `Dump()`, which should make it clear that it is not the same as `criu.Dump`. I am also yet to add comments that the LSP can use to provide documentation about a function. Adding that should further make it clear to users about what the function does.

rst0git: > I think you mean `Encode`, since `Dump` writes to the binary file. If we're renaming this, should we also rename `Load` to `Decode`? Or do we leave that as is?

IMHO yes, using `Encode` and `Decode` would make the code easier to understand.

crit/cmd/cli.go
```diff
@@ -0,0 +1,120 @@
...
+// The crit service used to invoke all commands
+var c crit.CritSvc
+
+// All members needed for crit struct
+var inputFilePath, outputFilePath, inputDirPath, exploreType string
+var pretty, noPayload bool
+
+var rootCmd = &cobra.Command{
+	Use:   "crit",
+	Short: "A tool to manipulate CRIU image files",
```
rst0git, line 23: What about `CRIU Image Tool`?

snprajwal: Makes sense :+1:

crit/crit.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	defer imgFile.Close()
+	// Convert binary image to Go struct
+	return countImg(imgFile)
+}
+
+func (c *crit) Parse() (*CriuImage, error) {
+	jsonData, err := os.ReadFile(c.inputFilePath)
+	if err != nil {
+		return nil, errors.New(fmt.Sprint("Error opening JSON file: ", err))
```
mihalicyn, line 72: opening -> reading

crit/decode.go
```diff
@@ -0,0 +1,270 @@
...
+func (img *CriuImage) decodeDefault(
+	f *os.File,
+	decodeExtra func(*os.File, proto.Message, bool) (string, error),
+	noPayload bool,
+) error {
+	sizeBuf := make([]byte, 4)
+	// Read payload size and payload until EOF
+	for {
+		n, err := f.Read(sizeBuf)
+		if n == 0 {
```
mihalicyn, line 85: It's better to check `err != nil` here

crit/decode.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+func (img *CriuImage) decodePagemap(f *os.File) error {
+	sizeBuf := make([]byte, 4)
+	// First entry is pagemap head
+	var payload proto.Message = &images.PagemapHead{}
+	// Read payload size and payload until EOF
+	for {
+		n, err := f.Read(sizeBuf)
+		if n == 0 {
```
mihalicyn, line 124: `err != nil`

crit/images/handler.go
```diff
@@ -0,0 +1,142 @@
...
+	case "FANOTIFY_FILE":
+		return &FanotifyFileEntry{}, nil
+	case "FANOTIFY_MARK":
+		return &FanotifyMarkEntry{}, nil
+	case "FDINFO":
+		return &FdinfoEntry{}, nil
+	case "FIFO":
+		return &FifoEntry{}, nil
+	case "FIFO_DATA":
+		return &FifoEntry{}, nil
```
mihalicyn, line 47: `&PipeDataEntry{}`?

crit/cmd/cli.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+	},
+}
+
+func init() {
+	// Disable completion generation
+	rootCmd.CompletionOptions.DisableDefaultCmd = true
+	// Decode options
+	decodeCmd.Flags().StringVarP(&inputFilePath, "input", "i", "",
+		"Path to the CRIU binary file")
```
mihalicyn, line 154: nit: `CRIU binary file` -> `CRIU image binary file`

mihalicyn:
Great work, Prajwal!

It would be great to have some ultra-simple integration test that will show that `go-crit` works equivalently to the original CRIT. I believe that we can start from some bash script and some image set generated from the ZDTM test suite run.

rst0git:
@snprajwal Would it be possible to include `crit` in the main Makefile as well?

Example:
```diff
diff --git a/Makefile b/Makefile
index 347062b..8ef29d0 100644
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ lint:
 build:
        $(MAKE) -C scripts
        $(GO) build -v ./...
+       $(MAKE) -C crit
```

crit/decode.go
```diff
@@ -0,0 +1,270 @@
...
+		return nil, err
+	}
+	magic := uint64(binary.LittleEndian.Uint32(magicBuf))
+	if magic == magicMap.ByName["IMG_COMMON"] ||
+		magic == magicMap.ByName["IMG_SERVICE"] {
+		if _, err := f.Read(magicBuf); err != nil {
+			return nil, err
+		}
+		magic = uint64(binary.LittleEndian.Uint32(magicBuf))
+	}
```
rst0git, line 34: The code for reading the magic in `decodeImg()` and `countImg()` looks very similar. Perhaps we could move it into a function (e.g., `readMagic()`) to make it reusable?

crit/decode.go
```diff
@@ -0,0 +1,270 @@
...
+				if _, err := f.Read(extraBuf); err != nil {
+					return err
+				}
+				entry.Extra = base64.StdEncoding.EncodeToString(extraBuf)
+			}
+			img.Entries = append(img.Entries, &entry)
+		}
+	} else {
+		if noPayload {
+			if _, err := f.Seek(0, 2); err != nil {
```
rst0git, line 200: It would be good to add a comment here. Why do we call `f.Seek(0, 2)`?

crit/decode-extra.go
```diff
@@ -0,0 +1,257 @@
...
+	"io"
+	"os"
+
+	"github.com/checkpoint-restore/go-criu/v5/crit/images"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
+
+// Helper to convert bytes into a more readable unit
+func countBytes(n int64) string {
```
rst0git, line 18: It might be better to move this function in a file called `utils.go`

crit/decode.go
snprajwal: The `(0, 2)` indicates "seek to an offset of 0 relative to the end of the file". https://pkg.go.dev/os#File.Seek
I've added a comment explaining this :+1:

crit/crit.go
snprajwal: `os.Readfile()` both opens and reads the entire contents of the file. Since we are reading all the way to the end, it's a cleaner solution than calling `os.Open()`, creating a buffer of the file size, and then using `f.Read()`.

crit/images/handler.go
snprajwal: Oops! Nice catch :)

snprajwal:
> It would be great to have some ultra-simple integration test that will show that go-crit works equivalently to the original CRIT. I believe that we can start from some bash script and some image set generated from the ZDTM test suite run.

@mihalicyn incoming in the next set of commits. I thought it would be better to complete `crit x` and be done with all the features before adding tests in.

crit/crit.go
mihalicyn: sure, I'm talking about error message text only. ;)

rst0git:
```go
errors.New(fmt.Sprint("Error reading JSON file: ", err))
```

PR #68, opened 2022-06-02
ci: add dependabot to update go mods

### LINKED ISSUE #67: Remove `vendor` dir and use `go.mod` directly
This issue is being opened to discuss the possibility of removing the `vendor` dir from the repo. Vendoring enables us to have reproducible builds, but at the cost of updates and security fixes in newer versions of the modules.

Including the upcoming `crit` library, we will be using three modules - `proto`, `x/sys` and `protojson`. All three are hardened packages with proper testing before every release. Go automatically fetches them when running `go build` based on the info in `go.mod`. We can use `go get -u` to update the deps manually when necessary.

snprajwal:
Fixes #67

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: monthly


adrianreber (approved):
LGTM

rst0git (approved):
LGTM

PR #71, opened 2022-06-21
Add makefile for testing

rst0git:
Running tests with CRIU requires root privileges, however, building go binaries does not. This patch separates the build steps from the steps used to run tests. This allows to build applications such as crit using `make` without root privileges.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,86 +1,24 @@
 SHELL = /bin/bash
 GO ?= go
 CC ?= gcc
-COVERAGE_PATH ?= $(shell pwd)/.coverage
-CRIU_FEATURE_MEM_TRACK = $(shell if criu check --feature mem_dirty_track > /dev/null; then echo 1; else echo 0; fi)
-CRIU_FEATURE_LAZY_PAGES = $(shell if criu check --feature uffd-noncoop > /dev/null; then echo 1; else echo 0; fi)
-CRIU_FEATURE_PIDFD_STORE = $(shell if criu check --feature pidfd_store > /dev/null; then echo 1; else echo 0; fi)
 
-export CRIU_FEATURE_MEM_TRACK CRIU_FEATURE_LAZY_PAGES CRIU_FEATURE_PIDFD_STORE
-
-all: build test phaul-test
+all: build
 
 lint:
 	golangci-lint run ./...
 
 build:
+	$(MAKE) -C scripts
 	$(GO) build -v ./...
 
-TEST_PAYLOAD := test/piggie/piggie
-TEST_BINARIES := test/test $(TEST_PAYLOAD) test/phaul/phaul
-COVERAGE_BINARIES := test/test.coverage test/phaul/phaul.coverage
-test-bin: $(TEST_BINARIES)
+test: build
+	$(MAKE) -C test
 
-test/piggie/piggie: test/piggie/piggie.c
-	$(CC) $^ -o $@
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,86 +1,24 @@
...
-export CRIU_FEATURE_MEM_TRACK CRIU_FEATURE_LAZY_PAGES CRIU_FEATURE_PIDFD_STORE
-
-all: build test phaul-test
+all: build
 
 lint:
 	golangci-lint run ./...
 
 build:
+	$(Q) $(MAKE) -C scripts
```
adrianreber, line 11: Do you define `$(Q)` somewhere? I don't think we have the setup in go-criu to support quiet builds.

Makefile
```diff
@@ -1,86 +1,24 @@
...
 CC ?= gcc
-COVERAGE_PATH ?= $(shell pwd)/.coverage
-CRIU_FEATURE_MEM_TRACK = $(shell if criu check --feature mem_dirty_track > /dev/null; then echo 1; else echo 0; fi)
-CRIU_FEATURE_LAZY_PAGES = $(shell if criu check --feature uffd-noncoop > /dev/null; then echo 1; else echo 0; fi)
-CRIU_FEATURE_PIDFD_STORE = $(shell if criu check --feature pidfd_store > /dev/null; then echo 1; else echo 0; fi)
 
-export CRIU_FEATURE_MEM_TRACK CRIU_FEATURE_LAZY_PAGES CRIU_FEATURE_PIDFD_STORE
-
-all: build test phaul-test
+all: build
```
snprajwal, line 5: Would we want to retain `test` here? Even though the test targets are moved into `test/Makefile`, we still have a `test` target to call it from this file.

Makefile
rst0git: Thank you for pointing this out. I have dropped `$(Q)` and updated the pull request.

Makefile
rst0git: > Would we want to retain test here?

Not really. This change allows to run `make` (without sudo) to execute only the `build` target.
We can use `sudo make test` to run tests as root.

test/Makefile
```diff
@@ -0,0 +1,82 @@
...
+codecov:
+	curl -Os https://uploader.codecov.io/latest/linux/codecov
+	chmod +x codecov
+	./codecov -f '.coverage/coverage.out'
+
+clean:
+	@rm -f $(TEST_BINARIES) $(COVERAGE_BINARIES) codecov
+	@rm -rf image $(COVERAGE_PATH)
+
+.PHONY: phaul-test test-bin clean coverage codecov
```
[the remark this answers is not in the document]
rst0git: `test`, `phaul/phaul` and `piggie/piggie` are not actually phony targets, but you are right about `all`.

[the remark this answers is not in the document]
snprajwal: I thought phony targets were those whose name does not match a file? Or is there more to it :eyes:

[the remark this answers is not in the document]
rst0git: Yes, you are right. For example, the following code would generate an executable file with the name `test`.
```Makefile
test: main.go
	$(GO) build -v -o $@ $^
```
`$@` is the name of the target:
https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html#Automatic-Variables

adrianreber (approved):

PR #73, opened 2022-06-24
test: drop test-bin target

rst0git:
The current [default target](https://www.gnu.org/software/make/manual/html_node/Rules.html) in test/Makefile is `test-bin` as this is the first target in the makefile. This was unintentional change introduced in https://github.com/checkpoint-restore/go-criu/pull/71. The default target should be `all`. However, `test-bin` does not appear to be used anywhere and thus, we can remove it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,16 +30,15 @@ jobs:
         go-version: ${{ matrix.go-version }}
 
     - name: Run tests
-      run: |
-        # Run actual test as root as it uses CRIU.
-        sudo make test
-        # Test magicgen script
-        make -C scripts test
+      run: sudo env "PATH=$PATH" make test
+
+    - name: Test magicgen script
+      run: make -C scripts test
 
     - name: Check code coverage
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -30,16 +30,15 @@ jobs:
         go-version: ${{ matrix.go-version }}
 
     - name: Run tests
-      run: |
-        # Run actual test as root as it uses CRIU.
-        sudo make test
-        # Test magicgen script
-        make -C scripts test
+      run: sudo env "PATH=$PATH:/usr/bin" make test
```
snprajwal, line 33: As far as I know, `PATH` for the root user is set to
`/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`
Are we sure we need to append it again?

rst0git: Thanks Prajwal, adding `/usr/bin` should not be necessary.

adrianreber (approved):

PR #75, opened 2022-07-11
magic-gen: refactor script with unit test

snprajwal:
The `magicgen.go` script was previously not testable, and has been
refactored to support tests. A unit test is also included, along with
the necessary changes in the Makefile.


rst0git (approved):
LGTM

snprajwal:
@adrianreber the old test is doing the same thing as the unit test, but with stdout and stdin instead. It seems kinda hacky, which is why I removed it. If you want, I can add in a proper integration test for this script along with the tests for #66

adrianreber:
> @adrianreber the old test is doing the same thing as the unit test, but with `stdout` and `stdin` instead. It seems kinda hacky, which is why I removed it. If you want, I can add in a proper integration test for this script along with the tests for #66

I think unit tests are important, but you are removing the test which tests the actual usage of the script. I think it is also important to test the script in that way it will be used. Removing an existing tests does not seem like a good approach. I you want to improve it later, no problem, but removing it now seems unnecessary.

PR #76, opened 2022-07-22
Add docs and tests for CRIT

snprajwal:
This PR provides docs and tests for the crit library introduced in #66 

- [X] Integration test
- [X] E2E test
- [X] Inline comments for LSP hover
- [ ] Documentation on CRIU website


crit/decode-extra.go
```diff
@@ -13,6 +13,7 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+// Extra data handler for pipe data
```
rst0git, line 16: Extra data handler for pipe/fifo data

crit/decode.go
```diff
@@ -121,6 +126,7 @@ func (img *CriuImage) decodePagemap(f *os.File) error {
 	return nil
 }
 
+// Special handler for ghost file
```
rst0git, line 129: *Special handler for ghost image?

crit/decode.go
```diff
@@ -92,6 +96,7 @@ func (img *CriuImage) decodeDefault(
 	return nil
 }
 
+// Special handler for pagemap file
```
rst0git, line 99: *Special handler for pagemap image?

crit/encode-extra.go
```diff
@@ -10,14 +10,17 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+// Extra data handler for pipe data
```
rst0git, line 13: Extra payload handler for pipe data?

crit/encode-extra.go
```diff
@@ -10,14 +10,17 @@ import (
...
 func encodePipesData(extra string) ([]byte, error) {
 	return base64.StdEncoding.DecodeString(extra)
 }
 
+// Extra data handler for socket queues
```
rst0git, line 18: Extra payload handler for socket queues?

crit/encode-extra.go
```diff
@@ -10,14 +10,17 @@ import (
...
 func encodeSkQueues(extra string) ([]byte, error) {
 	return base64.StdEncoding.DecodeString(extra)
 }
 
+// Extra data handler for TCP streams
```
rst0git, line 23: ditto

crit/encode-extra.go
```diff
@@ -36,10 +39,12 @@ func encodeTcpStream(extra string) ([]byte, error) {
 	return append(inqBytes, outQBytes...), nil
 }
 
+// Extra data handler for BPF map data
```
rst0git, line 42: ditto

crit/encode-extra.go
```diff
@@ -36,10 +39,12 @@ func encodeTcpStream(extra string) ([]byte, error) {
...
 func encodeBpfmapData(extra string) ([]byte, error) {
 	return base64.StdEncoding.DecodeString(extra)
 }
 
+// Extra data handler for IPC semaphores
```
rst0git, line 47: ditto

crit/encode-extra.go
```diff
@@ -62,6 +67,7 @@ func encodeIpcSem(extra string) ([]byte, error) {
 	return extraPayload, nil
 }
 
+// Extra data handler for IPC shared memory
```
rst0git, line 70: ditto

crit/encode-extra.go
```diff
@@ -75,6 +81,7 @@ func encodeIpcShm(extra string) ([]byte, error) {
 	return extraPayload, nil
 }
 
+// Extra data handler for IPC messages
```
rst0git, line 84: ditto

crit/encode.go
```diff
@@ -106,6 +110,7 @@ func (img *CriuImage) encodeDefault(
 	return nil
 }
 
+// Special handler for ghost file
```
rst0git, line 113: Special handler for ghost image

crit/explore.go
```diff
@@ -60,16 +62,20 @@ func (c *crit) ExplorePs() (*PsTree, error) {
 	return psTreeRoot, nil
 }
 
+// Fd represents the files opened by a single process
```
rst0git, line 65: Fd represents the file descriptors opened by a single process

crit/explore.go
```diff
@@ -60,16 +62,20 @@ func (c *crit) ExplorePs() (*PsTree, error) {
...
 	Files []*File `json:"files,omitempty"`
 }
 
+// File represents a single opened file
 type File struct {
 	Fd   string `json:"fd"`
 	Path string `json:"path"`
 }
 
+// ExploreFds searches the process tree for opened files
```
rst0git, line 77: *for open files

crit/explore.go
```diff
@@ -60,16 +62,20 @@ func (c *crit) ExplorePs() (*PsTree, error) {
...
+// and returns a list of process IDs and corresponding files
```
rst0git, line 78: * a list of PIDs along with corresponding files

crit/image.go
```diff
@@ -110,6 +117,7 @@ func unmarshalDefault(imgData *jsonImage, img *CriuImage) error {
 	return nil
 }
 
+// Special handler for ghost file
```
rst0git, line 120: *ghost image

crit/cmd/cli.go
```diff
@@ -35,7 +35,7 @@ var decodeCmd = &cobra.Command{
 	Long: `Convert the input binary image to JSON and write it to a file.
 If no output file is provided, the JSON is printed to stdout.`,
 	Run: func(cmd *cobra.Command, args []string) {
-		c = crit.New(inputFilePath, outputFilePath,
+		c = crit.NewCli(inputFilePath, outputFilePath,
```
rst0git, line 38: Would it be possible to move this commit in a separate pull request?

test/crit/crit-test.sh
```diff
@@ -0,0 +1,104 @@
...
+
+		echo "=== done"
+	done
+}
+
+
+function run_test2 {
+	mapfile -t array <<< "$images_list"
+
+	PROTO_IN=${array[0]}
```
rst0git, line 47: It looks like here we take the first image from `$images_list`. However, it might be easier to debug if we use a specific image instead (e.g., `inventory.img`).

test/crit/crit-test.sh
```diff
@@ -0,0 +1,104 @@
...
+function gen_imgs {
+	PID=$(./loop)
+	if ! criu dump -v4 -o dump.log -D ./ -t "$PID"; then
+		echo "Failed to checkpoint process $PID"
+		cat dump.log
+		kill -9 "$PID"
+		exit 1
+	fi
+
+	images_list=$(ls -1 ./*.img)
```
rst0git, line 16: Note that `crit decode` does not support `pages-X.img` files.

test/Makefile
```diff
@@ -12,7 +12,7 @@ TEST_PAYLOAD := piggie/piggie
 TEST_BINARIES := test $(TEST_PAYLOAD) phaul/phaul
 COVERAGE_BINARIES := test.coverage phaul/phaul.coverage
 
-all: $(TEST_BINARIES) phaul-test
+all: $(TEST_BINARIES) phaul-test crit-test
```
rst0git, line 15: `crit-test` generates a significant amount of logs.
IMHO it would be better to run them in a separate step in [`.github/workflows/main.yml`](https://github.com/checkpoint-restore/go-criu/blob/a6d7cb17ecc29ac8f9adc19e18f8c5193a416f1a/.github/workflows/main.yml):

```
sudo make -C test/ crit-test
```

test/crit/main.go
```diff
@@ -0,0 +1,83 @@
...
+		"route-",
+		"route6-",
+		"ifaddr-",
+		"tmpfs-",
+		"netns-ct-",
+		"netns-exp-",
+		"rule-",
+	}
+	var imgs []string
+	dir, err := os.ReadDir(".")
```
rst0git, line 65: Maybe we could use [`filepath.Walk()`](https://pkg.go.dev/path/filepath#WalkDir)?
```go
func getImgs() []string {
	var skipImgs = []string{
		"pages-",
		"iptables-",
		"ip6tables-",
		"nftables-",
		"route-",
		"route6-",
		"ifaddr-",
		"tmpfs-",
		"netns-ct-",
		"netns-exp-",
		"rule-",
	}
	var imgs []string

	re := regexp.MustCompile(`.*\.img`)
	fn := func(path string, info os.DirEntry, err error) error {
		switch {
		case err != nil:
			return err
		case !info.Type().IsRegular():
			return nil
		case !re.MatchString(info.Name()):
			return nil
		}

		imgName := info.Name()
		for _, skip := range skipImgs {
			if strings.HasPrefix(imgName, skip) {
				return nil
			}
		}

		imgs = append(imgs, path)
		return nil
	}

	err := filepath.WalkDir(".", fn)
	if err != nil {
		log.Fatal(err)
	}

	return imgs
}
```

test/crit/main.go
```diff
@@ -0,0 +1,83 @@
...
+		"ip6tables-",
+		"nftables-",
+		"route-",
+		"route6-",
+		"ifaddr-",
+		"tmpfs-",
+		"netns-ct-",
+		"netns-exp-",
+		"rule-",
+	}
```
rst0git, line 63: Should we include all raw images?

```bash
$ grep FD_ENTRY_F criu/image-desc.c 
#define FD_ENTRY_F(_name, _fmt, _f)     \
	FD_ENTRY_F(PIPES_DATA,	"pipes-data", O_NOBUF), /* splices data */
	FD_ENTRY_F(FIFO_DATA,	"fifo-data", O_NOBUF), /* the same */
	FD_ENTRY_F(SK_QUEUES,	"sk-queues", O_NOBUF), /* lseeks the image */
	FD_ENTRY_F(IPCNS_SHM,	"ipcns-shm-%u", O_NOBUF), /* writes segments of data */
	FD_ENTRY_F(GHOST_FILE,	"ghost-file-%x", O_NOBUF),
	FD_ENTRY_F(MEMFD_INODE,	"memfd", O_NOBUF),
	FD_ENTRY_F(IFADDR,	"ifaddr-%u", O_NOBUF),
	FD_ENTRY_F(ROUTE,	"route-%u", O_NOBUF),
	FD_ENTRY_F(ROUTE6,	"route6-%u", O_NOBUF),
	FD_ENTRY_F(RULE,	"rule-%u", O_NOBUF),
	FD_ENTRY_F(IPTABLES,	"iptables-%u", O_NOBUF),
	FD_ENTRY_F(IP6TABLES,	"ip6tables-%u", O_NOBUF),
	FD_ENTRY_F(NFTABLES,	"nftables-%u", O_NOBUF),
	FD_ENTRY_F(TMPFS_IMG,	"tmpfs-%u.tar.gz", O_NOBUF),
	FD_ENTRY_F(TMPFS_DEV,	"tmpfs-dev-%u.tar.gz", O_NOBUF),
	FD_ENTRY_F(AUTOFS,	"autofs-%u", O_NOBUF),
	FD_ENTRY_F(TTY_DATA,	"tty-data", O_NOBUF),
	FD_ENTRY_F(PAGES,	"pages-%u", O_NOBUF),
	FD_ENTRY_F(PAGES_OLD,	"pages-%d", O_NOBUF),
	FD_ENTRY_F(SHM_PAGES_OLD, "pages-shmem-%ld", O_NOBUF),
	FD_ENTRY_F(BPFMAP_FILE,	"bpfmap-file", O_NOBUF),
	FD_ENTRY_F(BPFMAP_DATA,	"bpfmap-data", O_NOBUF),
```

test/crit/Makefile
```diff
@@ -1,18 +1,24 @@
 CC ?= gcc
+GO ?= go
 
 all: test clean
 
 test: gen-imgs
+	@echo "Running integration test"
+	@./crit-test
```
rst0git, line 8: It might be good to separate these tests, e.g., `crit-test-integration` and `crit-test-e2e`.
This would make it easier to debug the tests.

test/crit/main.go
snprajwal: Since we know that all the generated images will be present at the top-level directory, `filepath.WalkDir()` seems like overkill. `os.ReadDir()` is also faster, so we can leverage that as we don't need to recurse subdirectories.

test/crit/main.go
snprajwal: `Decode()` and `Encode()` support some of these images, like ghost files, pipes data, etc. So I think we can retain those. I'll test which ones are not supported and add them in 👍

test/crit/main.go
```diff
@@ -0,0 +1,83 @@
...
+	dir, err := os.ReadDir(".")
+	if err != nil {
+		log.Fatal(err)
+	}
+	for _, file := range dir {
+		name := file.Name()
+		if filepath.Ext(name) == ".img" {
+			for _, skip := range skipImgs {
+				if strings.HasPrefix(name, skip) {
+					goto skipFile
```
rst0git, line 74: can we use `break` instead of `goto skipFile`?

test/crit/main.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		if !bytes.Equal(imgBytes, testImgBytes) {
+			log.Fatal("[RECODE]: files do not match")
+		}
+		log.Println("=== PASS")
+	}
+}
+
+func getImgs() []string {
+	var skipImgs = []string{
```
rst0git, line 51: It might be good to add a comment here that explains why do we skip these images.

test/crit/main.go
```diff
@@ -0,0 +1,83 @@
...
+		err = c.Encode(decodedImg)
+		if err != nil {
+			log.Fatal("[ENCODE]: ", err)
+			continue
+		}
+		// Open and compare original and test files
+		imgBytes, err := os.ReadFile(img)
+		if err != nil {
+			log.Fatal(err)
+			continue
```
rst0git, line 36: `log.Fatal()` would call `os.Exit(1)` and thus, `continue` is redundant?

test/crit/main.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		testImgBytes, err := os.ReadFile(testImg)
+		if err != nil {
+			log.Fatal(err)
+			continue
```
rst0git, line 41: ditto

test/crit/main.go
```diff
@@ -0,0 +1,83 @@
...
+	imgs := getImgs()
+	for _, img := range imgs {
+		log.Println("=== ", img)
+		testImg := fmt.Sprintf("%s.test.img", filepath.Base(img))
+		c := crit.New(img, testImg, "", false, false)
+		// Decode the binary image file
+		decodedImg, err := c.Decode()
+		if err != nil {
+			log.Fatal("[DECODE]: ", err)
+			continue
```
rst0git, line 24: `log.Fatal()` would call `os.Exit(1)`, and thus `continue` is redundant?

test/crit/main.go
snprajwal: `break` will get us out of the first for loop (which checks against the array of excluded patterns) but will still end up appending the image to the list in the outer for loop. Our options are either a flag or a label, and this one seemed cleaner. If you would like, I can change it to use a `bool skipFile = true` instead.

test/crit/main.go
snprajwal: Oops! I had added `continue` with `log.Println()` when I first wrote it, will remove 👍

test/crit/main.go
rst0git: Both [`break`](https://go.dev/doc/go1.17_spec#Break_statements) and [`continue`](https://go.dev/doc/go1.17_spec#Continue_statements) support labels but perhaps `continue` makes more sense in the current context:
```diff
--- a/test/crit/main.go
+++ b/test/crit/main.go
@@ -66,17 +66,18 @@ func getImgs() []string {
        if err != nil {
                log.Fatal(err)
        }
+
+nextFile:
        for _, file := range dir {
                name := file.Name()
                if filepath.Ext(name) == ".img" {
                        for _, skip := range skipImgs {
                                if strings.HasPrefix(name, skip) {
-                                       goto skipFile
+                                       continue nextFile
                                }
                        }
                        imgs = append(imgs, file.Name())
                }
-       skipFile:
        }
 
        return imgs
```

test/crit/main.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+}
+
+func getImgs() []string {
+	/* Certain image files generated by CRIU do not
+	use the protobuf format and contain raw binary
+	data. Some image files are also generated using
+	external tools (ifaddr, route, tmpfs). As these
+	images cannot be processed by CRIT, they are
+	excluded from the tests. */
```
rst0git, line 52: The convention for [variable comments](https://go.dev/doc/comment#var) is using `//`.

test/crit/main.go
```diff
@@ -0,0 +1,89 @@
...
+		c := crit.New(img, testImg, "", false, false)
+		// Decode the binary image file
+		decodedImg, err := c.Decode()
+		if err != nil {
+			log.Fatal("[DECODE]: ", err)
+		}
+		// Encode it into test binary image file
+		err = c.Encode(decodedImg)
+		if err != nil {
+			log.Fatal("[ENCODE]: ", err)
```
rst0git, line 28:
```suggestion
		if err = c.Encode(decodedImg); err != nil {
```

test/crit/Makefile
```diff
@@ -1,19 +1,29 @@
...
 all: test clean
 
-test: gen-imgs
+test: gen-imgs integration-test e2e-test
+
+integration-test:
+	@echo "Running integration test"
+	@./crit-test
+
+e2e-test:
```
rst0git, line 12: We need to add the corresponding dependency of targets to be able to run the tests with
`sudo make -C test/crit integration-test` and `sudo make -C test/crit e2e-test`
```diff
--- a/test/crit/Makefile
+++ b/test/crit/Makefile
@@ -3,17 +3,17 @@ GO ?= go
 
 all: test clean
 
-test: gen-imgs integration-test e2e-test
+test: integration-test e2e-test
 
-integration-test:
+integration-test: gen-imgs crit-test
        @echo "Running integration test"
        @./crit-test
 
-e2e-test:
+e2e-test: gen-imgs
        @echo "Running E2E test"
        @./crit-test.sh
 
-gen-imgs: ../loop/loop crit-test
+gen-imgs: ../loop/loop
        $(eval PID := $(shell ../loop/loop))
        criu dump -v4 -o dump.log -D ./ -t $(PID)
```

test/crit/main.go
```diff
@@ -0,0 +1,89 @@
...
+	var imgs []string
+	dir, err := os.ReadDir(".")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+nextFile:
+	for _, file := range dir {
+		name := file.Name()
+		if filepath.Ext(name) == ".img" {
```
rst0git, line 78: we might want to exclude `.json.img`, `.test.img` as well?

test/crit/Makefile
rst0git, line 12: We need to add the corresponding dependency targets to be able to run the tests with
`sudo make -C test/crit integration-test` and `sudo make -C test/crit e2e-test`
```diff
--- a/test/crit/Makefile
+++ b/test/crit/Makefile
@@ -3,17 +3,17 @@ GO ?= go
 
 all: test clean
 
-test: gen-imgs integration-test e2e-test
+test: integration-test e2e-test
 
-integration-test:
+integration-test: gen-imgs crit-test
        @echo "Running integration test"
        @./crit-test
 
-e2e-test:
+e2e-test: gen-imgs
        @echo "Running E2E test"
        @./crit-test.sh
 
-gen-imgs: ../loop/loop crit-test
+gen-imgs: ../loop/loop
        $(eval PID := $(shell ../loop/loop))
        criu dump -v4 -o dump.log -D ./ -t $(PID)
```

test/crit/main.go
snprajwal: I initially thought this would be more of a developer-specific comment, hence the multi line syntax

test/crit/main.go
snprajwal: I had ordered the tests in the makefile so that the integration test ran first, before the E2E test generated any of these files. Is there a possibility we'll run them in the wrong order?

test/crit/Makefile
snprajwal: Adding `gen-imgs` as a dep for the individual test targets would mean that we'll be generating two sets of images, which would make things messy. Since the individual targets will be called only for debugging, I thought we can just run `gen-imgs` manually before that.

test/crit/main.go
rst0git: Thank you for pointing this out. The idea is for the e2e and integration tests to be able to run independently.
For example, we can run the following:
```
sudo make -C test/crit/ e2e-test
sudo make -C test/crit/ integration-test
sudo make -C test/crit/ integration-test
```
In this case, we have to make sure that integration-test would not process the output files that have been generated from the previous run (`*.json.img`, `*.test.img`, as well as `tmp.*.img` directories).

test/crit/Makefile
rst0git: To solve this problem we can make this a non-phony target.
For example:
```diff
--- a/test/crit/Makefile
+++ b/test/crit/Makefile
@@ -3,19 +3,20 @@ GO ?= go
 
 all: test clean
 
-test: gen-imgs integration-test e2e-test
+test: integration-test e2e-test
 
-integration-test:
+integration-test: test-images crit-test
        @echo "Running integration test"
        @./crit-test
 
-e2e-test:
+e2e-test: test-images
        @echo "Running E2E test"
        @./crit-test.sh
 
-gen-imgs: ../loop/loop crit-test
+test-images: ../loop/loop
        $(eval PID := $(shell ../loop/loop))
-       criu dump -v4 -o dump.log -D ./ -t $(PID)
+       mkdir -p $@
+       criu dump -v4 -o dump.log -D $@ -t $(PID)
 
 ../loop/loop: ../loop/loop.c
        $(CC) $^ -o $@
@@ -24,6 +25,6 @@ crit-test: main.go
        $(GO) build -v -o $@ $^
 
 clean:
-       @rm -f *.img *.log *.json stats-*
+       @rm -rf test-images
 
-.PHONY: all test integration-test e2e-test gen-imgs clean
+.PHONY: all test integration-test e2e-test clean
```
It creates a folder called `test-images` and this target would run only when the folder does not exist.

test/crit/Makefile
```diff
@@ -0,0 +1,19 @@
...
+
+all: test clean
+
+test: gen-imgs
+	@echo "Running E2E test"
+	@./crit-test.sh
+
+gen-imgs: loop
+	$(eval PID := $(shell ./loop))
+	criu dump -v4 -o dump.log -D ./ -t $(PID)
```
rst0git: It might be good to use variable for the path to the `criu` binary: `CRIU ?= criu`
This conditional assignment would enable running the test with locally compiled criu instead.
```suggestion
	$(CRIU) dump -v4 -o dump.log -D ./ -t $(PID)
```

test/crit/crit-test.sh
```diff
@@ -0,0 +1,94 @@
...
+function run_test2 {
+	PROTO_IN=./inventory.img
+	JSON_IN=$(mktemp -p ./ tmp.XXXXXXXXXX.json)
+	OUT=$(mktemp -p ./ tmp.XXXXXXXXXX.img)
+
+	# prepare
+	${CRIT} decode -i "${PROTO_IN}" -o "${JSON_IN}"
+
+	# proto in - json out decode
+	cat "${PROTO_IN}" | ${CRIT} decode || exit 1
```
rst0git: > shellcheck test/crit/crit-test.sh
```
In test/crit/crit-test.sh line 46:
	cat "$PROTO_IN" | $CRIT decode || exit 1
            ^---------^ SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
```

snprajwal: We're doing that in line 52. The test just makes sure that the command works with every possible combination of input and output - redirection, pipes, and filenames with flags

rst0git: This is not really necessary, bash handles the input and output redirection:

> The output of each command in the pipeline is connected via a pipe to the input of the next command.

https://www.gnu.org/software/bash/manual/html_node/Pipelines.html

rst0git (approved):
LGTM

PR #77, opened 2022-07-28
fix(crit): read from stdin in CLI app

snprajwal:
The crit CLI app now supports reading input from stdin. Input can also
be sent through a pipe or I/O redirection.


rst0git (approved):
LGTM

PR #85, opened 2022-08-15
chore: bump Go version to 1.16

snprajwal:
1.16 ensures that we retain compatibility with Podman.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/checkpoint-restore/go-criu/v5
 
-go 1.13
+go 1.16
 
 require (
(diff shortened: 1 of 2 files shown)


rst0git:
@snprajwal We usually use the name of the file or directory that is changed as a prefix to the commit subject line.
https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#describe-your-changes

This makes it easier to understand the changes in every commit when reviewing the history with `git log --oneline`.
Would it be possible to use `go.mod: bump Go version to 1.16` instead?

rst0git (approved):
LGTM

Issue #86, opened 2022-08-15
v6.0.0 release


snprajwal:
## New features
- `crit` library
## Breaking changes
- Deprecate `stats` and `rpc` in favour of `crit`
## Related projects that will need PRs
- [x] [runc](https://github.com/opencontainers/runc/)
- [x] [checkpointctl](https://github.com/checkpoint-restore/checkpointctl)
- [x] [podman](https://github.com/containers/podman)
- [x] [LXD](https://github.com/lxc/lxd)

rst0git:
Prajwal and I discussed if we could deprecate the "stats" functionality of go-criu in favor of the [crit package](https://github.com/checkpoint-restore/go-criu/tree/master/crit).

This functionality is currently used for `--print-stats` in checkpointctl (https://github.com/checkpoint-restore/checkpointctl/pull/12) and Podman (https://github.com/containers/podman/pull/12184). However, to be able to update checkpointctl / Podman we might need to first create a new release of go-criu that includes the crit package.

In addition, somewhat similar functionality was introduced in LXD sometime ago that is parsing the `stats-dump` file to enable pre-copy migration (https://github.com/lxc/lxd/pull/4072). If we could integrate go-criu with LXD it would greatly simplify this part of the code and improve the maintainability.

@checkpoint-restore/go-criu-maintainers What do you think?

PR #87, opened 2022-08-17
deprecate: replace stats and rpc with crit

snprajwal:
The `crit` library allows users to decode all image files, including stats-dump and stats-restore, which is currently being done by the `stats` package. It also provides the proto bindings present in the `rpc` package. Hence, `stats` and `rpc` have been deprecated in favour of `crit` and will be removed in v6.0.0.


stats/utils.go
```diff
@@ -1,54 +0,0 @@
-package stats
```
rst0git, line 1: It might be better to add a comment that the `stats` package has been deprecated in favor of `crit` and will be removed in the next major release.

rst0git:
> I have not seen runc on the list so far. We should not forget that one.

Thanks for pointing this out Adrian. I have added runc to the list in https://github.com/checkpoint-restore/go-criu/issues/86.

rst0git (approved):
LGTM

PR #89, opened 2022-09-03
Add functions for stats

snprajwal:
This also fixes some of the places where `fmt.Sprintf` was being used instead of `filepath.Join`

### DIFF
--- a/crit/stats.go
+++ b/crit/stats.go
@@ -1,0 +1,46 @@
+package crit
+
+import (
+	"errors"
+	"path/filepath"
+
+	"github.com/checkpoint-restore/go-criu/v6/crit/images"
+)
+
+// Helper function to load stats file into Go struct
+func getStats(path string) (*images.StatsEntry, error) {
+	c := New(path, "", "", false, false)
+	statsImg, err := c.Decode()
+	if err != nil {
+		return nil, err
+	}
+
+	stats, ok := statsImg.Entries[0].Message.(*images.StatsEntry)
+	if !ok {
+		return nil, errors.New("Failed to type assert stats image")
+	}
+
+	return stats, nil
+}
+
+// GetDumpStats returns the dump statistics of a checkpoint.
+// dir is the path to the directory with the checkpoint images.
+func GetDumpStats(dir string) (*images.DumpStatsEntry, error) {
+	stats, err := getStats(filepath.Join(dir, "stats-dump"))
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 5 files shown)


crit/stats.go
```diff
@@ -0,0 +1,38 @@
...
+	c := New(filepath.Join(dir, "stats-restore"), "", "", false, false)
+	statsImg, err := c.Decode()
+	if err != nil {
+		return nil, err
+	}
+
+	stats, ok := statsImg.Entries[0].Message.(*images.StatsEntry)
+	if !ok {
+		return nil, errors.New("Failed to type assert stats-restore image")
+	}
```
rst0git, line 35: The code in these two functions looks very similar.
Would it be possible to move it into `getStats()` function that is then used by `GetDumpStats()` and `GetRestoreStats()`?

snprajwal: The file name changes for both functions (stats-dump and stats-restore). I feel like reworking it into a common function will actually make it worse to implement

rst0git: What about something like the following?
```go
func getStats(path string) (*images.StatsEntry, error) {
	c := New(path, "", "", false, false)
	statsImg, err := c.Decode()
	if err != nil {
		return nil, err
	}

	stats, ok := statsImg.Entries[0].Message.(*images.StatsEntry)
	if !ok {
		return nil, errors.New("Failed to type assert stats image")
	}

	return stats, nil
}

func GetDumpStats(dir string) (*images.DumpStatsEntry, error) {
	stats, err := getStats(filepath.Join(dir, "stats-dump"))
	if err != nil {
		return nil, err
	}

	return stats.GetDump(), nil
}

func GetRestoreStats(dir string) (*images.RestoreStatsEntry, error) {
	stats, err := getStats(filepath.Join(dir, "stats-restore"))
	if err != nil {
		return nil, err
	}

	return stats.GetRestore(), nil
}
```

snprajwal: Yep I landed at the same thing :+1:

rst0git:
@snprajwal It would be good if we can add a test case that can verify the functionality of `GetDumpStats()` and `GetRestoreStats()`. For example, something like https://github.com/rst0git/go-criu/commit/1ba7a011fc1624e12d181f7c223ef233170c598b

rst0git (approved):

adrianreber:
I would not include the README change mentioning 6.1.0. Besides that it looks good. Thanks. Especially as it includes tests.

PR #93, opened 2022-09-06
fix(crit/stats): add const for dump and restore

### LINKED ISSUE #92: const stats.StatsDump stats.StatsRestore missing
@snprajwal One of your previous PRs removed
```
package stats

const (
	StatsDump    = "stats-dump"
	StatsRestore = "stats-restore"
)
```

Can you re-create those constants and also use it in your stats wrapper functions. Those constants are used in CRI-O and Podman.

@rst0git this would require a new release. 6.2.0 probably

snprajwal:
Closes #92

### DIFF
--- a/crit/stats.go
+++ b/crit/stats.go
@@ -7,6 +7,11 @@ import (
 	"github.com/checkpoint-restore/go-criu/v6/crit/images"
 )
 
+const (
+	StatsDump    = "stats-dump"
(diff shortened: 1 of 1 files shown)


adrianreber (approved):
Looks good to me.

@rst0git if you agree, please also make a 6.2.0 release.

PR #95, opened 2022-09-07
crit/images: fix name conflict with letsencrypt

### LINKED ISSUE #94: proto: namespace conflict with letsencrypt
Unfortunately https://github.com/checkpoint-restore/go-criu/pull/90 did not resolve the problem with conflicting proto file names. Since protobuf v1.26.0 a hard error will be reported when the compiled Go program starts up that has multiple conflicting protobuf names linked into it.

**Steps to reproduce the issue:**
1. `git clone -b go-criu/v6 git@github.com:rst0git/libpod.git`
2. ```cd libpod && make -j `nproc` && sudo make test```

**Describe the results you received:**
The tests fail with the following error:
```
panic: proto: file "sa.proto" is already registered
	previously from: "github.com/letsencrypt/boulder/sa/proto"
	currently from:  "github.com/checkpoint-restore/go-criu/v6/crit/images"
See https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict


goroutine 1 [running]:
google.golang.org/protobuf/reflect/protoregistry.glob..func1({0x1b0b380?, 0xc0004ce240?}, {0x1b0b380?, 0xc0004ce280})
	/home/rst/tmp/libpod/vendor/google.golang.org/protobuf/reflect/protoregistry/registry.go:56 +0x1ee
google.golang.org/protobuf/reflect/protoregistry.(*Files).RegisterFile(0xc0001a8678, {0x1b204c8?, 0xc0000fc000?})
	/home/rst/tmp/libpod/vendor/google.golang.org/protobuf/reflect/protoregistry/registry.go:130 +0x39c
google.golang.org/protobuf/internal/filedesc.Builder.Build({{0x153c2e0, 0x34}, {0x25a1a80, 0x100, 0x100}, 0x0, 0x1, 0x0, 0x0, {0x1b10fd0, ...}, ...})
	/home/rst/tmp/libpod/vendor/google.golang.org/protobuf/internal/filedesc/build.go:112 +0x1d6
google.golang.org/protobuf/internal/filetype.Builder.Build({{{0x153c2e0, 0x34}, {0x25a1a80, 0x100, 0x100}, 0x0, 0x1, 0x0, 0x0, {0x0, ...}, ...}, ...})
	/home/rst/tmp/libpod/vendor/google.golang.org/protobuf/internal/filetype/build.go:138 +0x19d
github.com/checkpoint-restore/go-criu/v6/crit/images.file_sa_proto_init()
	/home/rst/tmp/libpod/vendor/github.com/checkpoint-restore/go-criu/v6/crit/images/sa.pb.go:190 +0x198
github.com/checkpoint-restore/go-criu/v6/crit/images.file_core_proto_init()
	/home/rst/tmp/libpod/vendor/github.com/checkpoint-restore/go-criu/v6/crit/images/core.pb.go:1109 +0x6f
github.com/checkpoint-restore/go-criu/v6/crit/images.init.12()
	/home/rst/tmp/libpod/vendor/github.com/checkpoint-restore/go-criu/v6/crit/images/core.pb.go:1095 +0x17
path is /home/rst/tmp/libpod/.coverage
Unable to read coverage file cmd/podman/coverprofile to combine, open cmd/podman/coverprofile: no such file or directory

Ginkgo ran 1 suite in 2m10.452270337s
Test Suite Failed
```

I can confirm that the initial solution renaming `core.proto` / `so.proto` files to `criu-core.proto` / `criu-so.proto` works as expected.

rst0git:
The core.proto and so.proto files provided by criu have conflicting names with the proto files included in https://github.com/letsencrypt/boulder. Since protobuf v1.26.0 a hard error is reported when a compiled Go program starts up that has multiple conflicting protobuf names linked into it. This pull request fixes this error by prepending the `criu-` prefix to `core.proto` and `so.proto`.

Fixes #94


crit/Makefile
```diff
@@ -22,9 +22,13 @@ update-proto:
...
 	# now registered under the package name `criu`.
 	ls ./images/*.proto | xargs sed -i '/syntax/a package criu;'
+	mv ./images/sa.proto ./images/criu-sa.proto
+	sed -i 's/sa\.proto/criu-sa\.proto/g' images/*.proto
+	mv ./images/core.proto ./images/criu-core.proto
+	sed -i 's/core\.proto/criu-core\.proto/g' images/*.proto
 
 gen-proto:
-	rm ./images/*.pb.go || true
+	rm -f ./images/*.pb.go || true
```
snprajwal, line 31: Will we need both `-f` and `|| true`?

rst0git: Thank you for pointing this out! I've dropped this change.

rst0git:
Steps to test this patch:
```
git clone -b go-criu/v6 https://github.com/rst0git/libpod
cd libpod && rm -rf vendor/github.com/checkpoint-restore/go-criu/v6
git clone -b fix-proto-name-conflict https://github.com/rst0git/go-criu/ vendor/github.com/checkpoint-restore/go-criu/v6
make -j `nproc`
sudo make test
```

crit/Makefile
```diff
@@ -22,6 +22,10 @@ update-proto:
 	# from other projects, the files of this repository are
 	# now registered under the package name `criu`.
 	ls ./images/*.proto | xargs sed -i '/syntax/a package criu;'
```
snprajwal, line 24: I think we should revert this change, since it introduces another layer of inconsistency with the original images from the criu repo. If our only solution is to rename the files to `criu-sa.proto` and `criu-core.proto`, then the package directive does not seem to benefit us in any way. We can either revert the changes in this PR, or revert the commit introduced in #90 before releasing v6.2.0. What do you think?

adrianreber: Yes, if the package `criu` does not gain us anything it is not necessary and just increases the difference between `go-criu` and `criu`.

Related to that, a bit, I am always a bit unhappy when I see that importing `go-criu` requires that all `*proto` files are imported even if unneeded. Is there a way to have each file in its own directory. That probably breaks `import` all over the place. As we are renaming files and changing `import` statements already... But that would be to big a change.

The thing I don't like about this change is that we need to rename files every time  a new collision happens and so far none of the projects importing `go-criu` currently need `sa.proto` and `core.proto`. It would be good to have an abstraction in Go which does only import the files needed and not everything. So far none of the tools using `go-criu` includes the phaul directory as it is not imported in the Go code. Can we make a change to image files in such a way that only the needed files are imported and not everything. Because that leads to the problems we currently have.

snprajwal: > It would be good to have an abstraction in Go which does only import the files needed and not everything

@adrianreber Unfortunately, Go imports all files that may _potentially_ be involved in the package. Since `crit/images/handler.go` has all the image types in the switch cases, Go will include all of the `.pb.go` files, and by extension all of the `.proto` files. I too dislike this form of package management, but I do not think there's any workaround we can do from our side :(

adrianreber: At this point it feels we tried everything and nothing really helps us.

Creating a subdirectory for each `*proto` file also does not work as we have `import` in the `*proto` files we would need to be adapted.

The best solution would be that CRIU renames all files to `criu-*`. But this also sounds difficult to get merged. Maybe it is worth a try at some point. Anyway... Sounds like we have to merge it the way it is now.

adrianreber:
@rst0git I do not see the result of your revert. How I understand GitHub all the files still have `package criu` in them.

rst0git:
> How I understand GitHub all the files still have `package criu` in them.

Thank you Adrian for pointing this out! There was some conflict with the revert changes. I have separated the revert changes in a separate commit to resolve this issue.

snprajwal (approved):
LGTM

rst0git:
@adrianreber PTAL

adrianreber:
Just wanted to make sure this is ready before merging. Is it ready?

rst0git:
> Just wanted to make sure this is ready before merging. Is it ready?

Yes, the first commit reverts the change introducing the package directive (as suggested in the [code review](https://github.com/checkpoint-restore/go-criu/pull/95#discussion_r964530499)). The second commit renames the conflicting `core.proto` and `so.proto` files to enable integration of go-criu with Podman (as we currently don't have a better solution to this problem).

I have tested the changes in this pull request locally (following the [steps described above](https://github.com/checkpoint-restore/go-criu/pull/95#issuecomment-1238929308)) and can confirm that it should resolve the CI errors we currently see in https://github.com/containers/podman/pull/15591.

adrianreber (approved):

PR #97, opened 2022-10-06
Optimize stats and rpc

rst0git:
We recently introduced support for CRIT in go-criu. However, the linker in the Go toolchain creates statically-linked binaries, including the Go runtime and other information necessary to support dynamic type checks. As a result, when updating Podman to use the latest go-criu version, the auto-generated protobuf binding for all CRIU images are compiled into the Podman binary, significantly increasing its size.

This pull request partially reverts the changes in https://github.com/checkpoint-restore/go-criu/pull/87 to allow us to upgrade go-criu without increasing the size of the compiled binaries.


rst0git:
cc @snprajwal

adrianreber (approved):

PR #98, opened 2022-10-12
crit: remove stats image

snprajwal:
Since stats functionality is provided in a separate package, `stats.proto` and `stats.pb.go` can safely be removed from `crit`.


snprajwal:
@rst0git PTAL

crit/images/handler.go
```diff
@@ -113,8 +113,6 @@ func ProtoHandler(magic string) (proto.Message, error) {
 		return &SignalfdEntry{}, nil
 	case "SK_QUEUES":
 		return &SkPacketEntry{}, nil
-	case "STATS":
-		return &StatsEntry{}, nil
```
rst0git, line 117: Hi @snprajwal, it would be good to keep the Go version of CRIT compatible with the Python version.
I think we can refactor crit to allow images to be imported as separate packages to avoid the problem with increased binary size.

PR #103, opened 2022-11-12
ci: add codeql as suggested by the bot (and fixes)

adrianreber:
This version of the PR from the bot also includes the corresponding fixes.

Same as #102 but with the corresponding fixes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,14 +4,12 @@ on: [push, pull_request]
 
 jobs:
   test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
         go-version: [1.17.x, 1.18.x, 1.19.x]
-        # criu is pre-installed in the Ubuntu 20.04 GitHub Action virtual
-        # environment as a dependency for Podman
-        criu_branch: ["", criu-dev]
+        criu_branch: [master, criu-dev]
 
     steps:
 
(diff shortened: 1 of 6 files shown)


adrianreber:
@rst0git PTAL for codeql integration

crit/Makefile
```diff
@@ -10,12 +10,13 @@ proto_files := $(sort $(subst $(proto_path)/,,$(wildcard $(proto_path)/*.proto))
 # and replace all spaces with commas to use with go_opt flag
 comma := ,
 proto_opts := $(subst $() $(),$(comma),$(patsubst %,M%=$(import_path),$(proto_files)))
+branch := master
```
rst0git, line 13: Should we use `GIT_BRANCH` (upper case) for this constant instead?

.github/workflows/main.yml
```diff
@@ -38,7 +38,13 @@ jobs:
       run: make -C scripts test
 
     - name: Test crit
-      run: sudo make -C test crit-test
+      run: |
+        make -C crit update-proto branch=${{ matrix.criu_branch }}
+        # If the definition of the protobuf files has changed upstream
+        # we just skip this test. Changed protobuf files will mean
+        # that crit-test will fail.
+        if [[ $(git status --porcelain | wc -l) -gt 0 ]]; then echo "Skipping crit-test because criu upstream changed protobuf definitions"; exit 0; fi
```
rst0git, line 46: Hmm, I'm not sure if this is a good idea.

We could just run `make gen-proto` and `crit-test` would pass even when the protobuf files have changed in the upstream repository:
```bash
make -C crit update-proto branch=${{ matrix.criu_branch }}
make -C crit gen-proto
sudo make -C test crit-test
```

adrianreber: Good idea

adrianreber: It was a good idea, but the protobuf version in Ubuntu 20.04 and 22.04 is too old to work with the files we generated with newer protobuf versions. Need to figure out how to best solve this.

adrianreber:
Looks like a found a way to make it all work. @rst0git PTAL

rst0git (approved):
LGTM

PR #104, opened 2022-11-15
Remove execute bit from source file

gibmat:
While working on updating the Debian packaging of this library, I noticed that crit/images/core-mips.proto needlessly has the execute bit set.


adrianreber (approved):

adrianreber:
@gibmat Thanks for your fix. Could you also submit the same thing to the main criu repository. We import the protobuf definitions from there and this will re-introduce the execute bit once we do that.

PR #108, opened 2023-03-26
ci: disable musttag in golangci-lint

snprajwal:
The musttag checker exits with status code 1 and kills the linter pipeline. Since nothing else is logged, disabling it seems to be the only fix for now. Running golangci-lint locally does not cause this issue, so it looks like a problem with the specific version of `musttag` that's being used in the GitHub action. I'll ping the golangci folks and let them know about this.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,6 +4,8 @@ linters:
     - performance
     - unused
     - format
+  disable:
+    - musttag
 
 linters-settings:
   exhaustive:


snprajwal:
@rst0git PTAL

adrianreber:
Can you provide the original error message and a link to the upstream report?

snprajwal:
Here's the logs that golangci-lint throws (I ran the action in both normal and debug mode, same logs are produced)

```
  Running [/home/runner/golangci-lint-1.52.1-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
  level=warning msg="[runner] Can't run linter goanalysis_metalinter: musttag: running `go list all`: exit status 1"
  level=error msg="Running error: 1 error occurred:\n\t* can't run linter goanalysis_metalinter: musttag: running `go list all`: exit status 1\n\n"
  
  Error: golangci-lint exit with code 3
```

I've opened an issue [here](https://github.com/golangci/golangci-lint-action/issues/714)

adrianreber:
Thanks. Let's merge this for now and if you ever get feedback from your issue we can re-enable it if we want.

adrianreber (approved):

PR #111, opened 2023-03-27
ci: add markdown linter

snprajwal:
This introduces a markdown linter for the go-criu repo. It runs only if a PR modifies any of the README.md files present. The commit also fixes the initial issues raised on the top-level readme.

cc @adrianreber

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -3,7 +3,7 @@ on: [ push, pull_request ]
 
 jobs:
   lint:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
     - uses: golangci/golangci-lint-action@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/verify.yml
```diff
@@ -20,3 +20,22 @@ jobs:
       run: |
         go mod tidy
         git diff --exit-code
+
+  lint_markdown:
+    runs-on: ubuntu-20.04
```
adrianreber, line 25: This could be `latest` instead of `20.04`.

snprajwal: The other two jobs were tagging 20.04, so I decided to stick with that for this one too. Would you prefer I change all of them to `latest` instead?

README.md
```diff
@@ -1,18 +1,22 @@
+<!-- markdownlint-disable MD010 -->
```
adrianreber, line 1: https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md010---hard-tabs

What about using spaces instead of tabs and keeping the rule active

adrianreber:
Thanks!

.github/workflows/verify.yml
adrianreber: You can try 22.04 or latest. If it doesn't break anything, we can switch. For a newly introduced check we can go with latest.

README.md
snprajwal: Got the config working :)

adrianreber (approved):

PR #117, opened 2023-04-13
Adapt code coverage to Go 1.20

adrianreber:
This will not work with older versions of Go.

This also introduces a commit to actually run the tests during CI. `test` -> `./test`.


adrianreber:
Now coverage works again, but the coverage is really low. Hmm. Ah, I might need to exclude certain packages.

test/Makefile
```diff
@@ -41,41 +41,38 @@ crit-test:
 	$(MAKE) -C crit
 
 test.coverage: *.go
-	$(GO) test \
-		-covermode=count \
-		-coverpkg=./... \
-		-mod=vendor \
-		-tags coverage \
-		-buildmode=pie -c -o $@ $^
+	# This requires at least Go 1.20
```
rst0git, line 44: Would it be better to use the following to check the Go version?

```Makefile
GO_MAJOR_VER = $(shell $(GO) version | cut -c 14- | cut -d' ' -f1 | cut -d'.' -f1)
GO_MINOR_VER = $(shell $(GO) version | cut -c 14- | cut -d' ' -f1 | cut -d'.' -f2)
MIN_GO_MAJOR_VER = 1
MIN_GO_MINOR_VER = 20
GO_VALIDATION_ERR = Go version is not supported. Please update to at least $(MIN_GO_MAJOR_VER).$(MIN_GO_MINOR_VER)

check-go-version:
	@if [ $(GO_MAJOR_VER) -gt $(MIN_GO_MAJOR_VER) ]; then \
		exit 0 ;\
	elif [ $(GO_MAJOR_VER) -lt $(MIN_GO_MAJOR_VER) ]; then \
		echo '$(GO_VALIDATION_ERR)';\
		exit 1; \
	elif [ $(GO_MINOR_VER) -lt $(MIN_GO_MINOR_VER) ] ; then \
		echo '$(GO_VALIDATION_ERR)';\
		exit 1; \
	fi

.PHONY: check-go-version
```
```Makefile
coverage: check-go-version $(COVERAGE_BINARIES) $(TEST_PAYLOAD)
```

adrianreber:
image

The problem was that the whole setup was broken. We were only collecting coverage from the `test/` directory. So the code coverage setup is correct.

test/Makefile
adrianreber: good idea, will do that

adrianreber: done

adrianreber:
I think this is ready now. The drop in test coverage is correct, because it was just plainly wrong what was reported before. Now we get the test coverage from all our code before we just got the test coverage from the test binaries (which is a pretty useless metric).

PR #118, opened 2023-04-14
Fix crash when feature check environment variables are not set

adrianreber:
This PR mainly fixes a crash when the feature check environment variables were not set.

I included another commit that fixes a complaint from `staticcheck` and enabled in a third commit additional CI checks.


rst0git (approved):
LGTM

PR #121, opened 2023-04-15
test: enable coverage of crit unit/e2e tests

rst0git:
This pull request enables code coverage for CRIT's [unit](https://go.dev/blog/cover) and [e2e](https://go.dev/testing/coverage/) tests.

### DIFF
--- a/crit/Makefile
+++ b/crit/Makefile
@@ -40,8 +40,8 @@ gen-proto:
 bin/crit: cmd/cli.go
 	$(GO) build ${GOFLAGS} -o $@ $^
 
-../test/loop/loop: ../test/loop/loop.c
-	$(CC) $^ -o $@
+../test/loop/loop:
+	$(MAKE) -C ../test/loop
 
 test-imgs: ../test/loop/loop
 	$(eval PID := $(shell ../test/loop/loop))
--- a/crit/Makefile
+++ b/crit/Makefile
@@ -51,6 +51,7 @@ test-imgs: ../test/loop/loop
 	pkill -9 loop
 
 unit-test: test-imgs
+	$(eval GOFLAGS ?= -cover)
 	go test ${GOFLAGS} -v ./...
 
 clean:
--- a/crit/Makefile
+++ b/crit/Makefile
@@ -1,4 +1,5 @@
 GO ?= go
+CRIU ?= criu
 
 # The import path that protoc will use if a proto file imports another one
 import_path := github.com/checkpoint-restore/go-criu/crit/images
@@ -39,7 +40,21 @@ gen-proto:
 bin/crit: cmd/cli.go
 	$(GO) build -o $@ $^
 
-clean:
-	rm -f bin/crit
+../test/loop/loop: ../test/loop/loop.c
+	$(CC) $^ -o $@
 
-.PHONY: all gen-proto update-proto clean
+test-imgs: ../test/loop/loop
+	$(eval PID := $(shell ../test/loop/loop))
+	mkdir -p $@
+	$(CRIU) dump -v4 -o dump.log -D $@ -t $(PID)
+	$(CRIU) restore -v4 -o restore.log -D $@ -d
+	pkill -9 loop
+
+unit-test: test-imgs
+	go test ${GOFLAGS} -v ./...
+
+clean:
+	@rm -f bin/crit
+	@rm -rf test-imgs
+
+.PHONY: all gen-proto update-proto unit-test clean
--- a/crit/Makefile
+++ b/crit/Makefile
@@ -38,7 +38,7 @@ gen-proto:
 		$(proto_files)
 
 bin/crit: cmd/cli.go
-	$(GO) build -o $@ $^
+	$(GO) build ${GOFLAGS} -o $@ $^
 
 ../test/loop/loop: ../test/loop/loop.c
 	$(CC) $^ -o $@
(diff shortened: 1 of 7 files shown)


rst0git:
@snprajwal PTAL

crit/Makefile
```diff
@@ -37,9 +38,23 @@ gen-proto:
 		$(proto_files)
 
 bin/crit: cmd/cli.go
-	$(GO) build -o $@ $^
+	$(GO) build ${GOFLAGS} -o $@ $^
+
+../test/loop/loop:
```
adrianreber, line 43: This is something I saw already earlier. We now have two test cases. piggie and loop. Is there a big difference? Maybe something for a future PR. Use the same thing for everything.

test/Makefile
```diff
@@ -88,10 +89,14 @@ coverage: check-go-version $(COVERAGE_BINARIES) $(TEST_PAYLOAD)
 	pkill -9 piggie; \
 	}
 	cd crit/ && GOCOVERDIR=${COVERAGE_PATH} ./crit-test.coverage
+	$(MAKE) -C ../crit/ unit-test GOFLAGS="-coverprofile=${COVERAGE_PATH}/coverprofile-crit-unit-test"
+	$(MAKE) -C crit/ e2e-test GOCOVERDIR=${COVERAGE_PATH}
 	$(MAKE) -C crit/ clean
 	# Print coverage from this run
 	$(GO) tool covdata percent -i=${COVERAGE_PATH}
 	$(GO) tool covdata textfmt -i=${COVERAGE_PATH} -o ${COVERAGE_PATH}/coverage.out
+	cat .coverage/coverprofile* | \
```
adrianreber, line 98: Just a question if this is necessary with Go 1.20? Seeing this seems to imply it is necessary, but just asking to be sure if this is really necessary.

adrianreber:
Nice change. Moving the unit test to `crit/` sounds like the right thing to do.

adrianreber:
```
go test  -v ./...
=== RUN   TestGetDumpStats
--- PASS: TestGetDumpStats (0.00s)
=== RUN   TestGetRestoreStats
--- PASS: TestGetRestoreStats (0.00s)
PASS
ok  	github.com/checkpoint-restore/go-criu/v6/crit	0.004s
?   	github.com/checkpoint-restore/go-criu/v6/crit/cmd	[no test files]
?   	github.com/checkpoint-restore/go-criu/v6/crit/images	[no test files]
```
This is also a good indicator that we should add some more tests.

snprajwal:
@adrianreber `crit/images` and `crit/cmd` are being refactored in the other PR, I'll add the tests there for the parts we currently aren't testing

adrianreber:
> @adrianreber `crit/images` and `crit/cmd` are being refactored in the other PR, I'll add the tests there for the parts we currently aren't testing

Sounds good.

adrianreber:
One thing I am thinking about is that for unit tests we could always generate test coverage output while using `go test`. Just to reduce code in the Makefiles. @rst0git what do you think? We can also merge what we have in this PR. Works both for me.

snprajwal:
We can have a `unit-test` target in the top level Makefile (or the Makefile in `test`) which simply runs `go -coverprofile=$COVER_PROFILE_PATH/coverprofile-unit-test test ./...` in the root.

adrianreber:
> We can have a `unit-test` target in the top level Makefile (or the Makefile in `test`) which simply runs `go -coverprofile=$COVER_PROFILE_PATH/coverprofile-unit-test test ./...` in the root.

Sounds like a good idea.

test/Makefile
rst0git: Yes, the test coverage from `go test` is stored in `coverprofile-crit-unit-test` but with `codecov` we upload only `coverage.out`. This is necessary for the coverage results to show up in GitHub.

crit/Makefile
rst0git: Sure. Let's fix this in another pull request when we merge @snprajwal's changes from https://github.com/checkpoint-restore/go-criu/pull/109.

snprajwal (approved):
LGTM

adrianreber:
@rst0git works for me. There still seems to be one question open. What about always running unit tests with test coverage results? Unit test coverage does not depend on Go 1.20 and can always be created.

rst0git:
> What about always running unit tests with test coverage results? Unit test coverage does not depend on Go 1.20 and can always be created.

I have added a commit to always run the unit tests with test coverage results.

crit/Makefile
```diff
@@ -37,9 +38,24 @@ gen-proto:
...
+
+test-imgs: ../test/loop/loop
+	$(eval PID := $(shell ../test/loop/loop))
+	mkdir -p $@
+	$(CRIU) dump -v4 -o dump.log -D $@ -t $(PID)
+	$(CRIU) restore -v4 -o restore.log -D $@ -d
+	pkill -9 loop
+
+unit-test: test-imgs
+	$(eval GOFLAGS ?= -cover)
```
adrianreber, line 54: How does `eval` work? That is new to me.

rst0git: > it allows you to define new makefile constructs that are not constant; which are the result of evaluating other variables and functions.

https://www.gnu.org/software/make/manual/html_node/Eval-Function.html

rst0git: `GOFLAGS` will be defined by default only in `unit-test`, but not in `bin/crit`.

adrianreber (approved):

PR #125, opened 2023-04-16
Add top-level clean make target

adrianreber:
And propagate clean to all directories.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,4 +38,8 @@ vendor:
 	$(GO) mod vendor
 	$(GO) mod verify
 
-.PHONY: build test lint vendor coverage codecov
+clean:
+	$(MAKE) -C crit/ clean
+	$(MAKE) -C test/ clean
+
+.PHONY: build test lint vendor coverage codecov clean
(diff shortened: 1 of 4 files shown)


test/loop/Makefile
```diff
@@ -2,3 +2,8 @@ CC ?= gcc
 
 loop: loop.c
 	$(CC) $^ -o $@
+
+clean:
+	rm -f loop
```
snprajwal, line 7: I think we would want to silence the command itself, like in the other Makefiles
```suggestion
	@rm -f loop
```

Makefile
```diff
@@ -38,4 +38,8 @@ vendor:
 	$(GO) mod vendor
 	$(GO) mod verify
 
-.PHONY: build test lint vendor coverage codecov
+clean:
+	make -C crit/ clean
+	make -C test/ clean
```
rst0git, line 43: we should use `$(MAKE)` for [recursive invocations](https://www.gnu.org/software/make/manual/html_node/Recursion.html): https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html#MAKE-Variable

adrianreber:
Applied changes as suggested.

rst0git (approved):
LGTM

PR #128, opened 2023-04-18
ci: add `stylecheck` to golangci-lint

### LINKED ISSUE #127: Enable 'stylecheck' golangci-lint linter
Trying to enable the golangci-lint linter 'stylecheck' I see a couple of reports in `crit/`.

@snprajwal maybe you could enable the linter and have a look at the reports and fix it if appropriate.

snprajwal:
Fixes #127


snprajwal:
PTAL @adrianreber

adrianreber:
Thanks.

Although most changes are done to internal functions this technically means this might break API. I am not aware of any external user of crit so far, so not sure if we need to increase the version to v7 for go-criu or not. I think `PsTree` might be the only structure which would count as API change. I think we do not need v7, but maybe we need it anyway at some point in the future. Just wanted to mention it here.

adrianreber (approved):

PR #131, opened 2023-05-11
Improve tests, code coverage, and Makefile targets

snprajwal:
- The CRIT E2E test did not cover the `crit info` and `crit show` commands. The test has been extended to include these.
- The code coverage for autogenerated files is not required to be reported. The Makefile target has been modified to skip these files before uploading the coverage report.
- Due to enclosing the test command in a block, shell would return exit code 0 even if the command running the test failed. This has been fixed using a trap to perform the cleanup instead.

PTAL @adrianreber @rst0git


test/Makefile
```diff
@@ -20,11 +20,10 @@ GO_VALIDATION_ERR = Go version is not supported. Please update to at least $(MIN
 
 all: $(TEST_BINARIES) phaul-test
 	mkdir -p image
-	PID=$$(piggie/piggie) && { \
+	PID=$$(piggie/piggie) && \
+	trap 'pkill -9 piggie' INT EXIT TERM && \
```
adrianreber, line 24: Ah, just had a look here. Instead of doing a pkill we could also do a kill on $PID, right? Less ambiguous. Instead of killing all processes with that name we would just kill the one.

snprajwal: Nice catch!

adrianreber (approved):
Looks good

test/Makefile
```diff
@@ -20,11 +20,10 @@ GO_VALIDATION_ERR = Go version is not supported. Please update to at least $(MIN
 
 all: $(TEST_BINARIES) phaul-test
 	mkdir -p image
-	PID=$$(piggie/piggie) && { \
+	PID=$$(piggie/piggie) && \
+	trap "kill -9 $$PID" INT EXIT TERM && \
```
rst0git, line 24:
```Suggestion
	trap "pkill -9 piggie" INT EXIT TERM && \
```
When running `make test` it shows the following:

> /bin/bash: line 1: kill: (22507) - No such process

adrianreber: Right. piggie creates a PID namespace and therefore the PID changes after restore. The original code using pkill is probably needed.

snprajwal: Agh I forgot to run the tests after amending the commit. Fixing this ASAP

rst0git (approved):
LGTM

PR #133, opened 2023-06-02
crit: add features to retrieve memory pages

behouba:
The goal of this PR is to add features for collecting memory pages content. 
For the now I have added `GetMemPages` function to retrieve the memory pages of a process. Alongside it, helper functions `genMemChunk` and `getPage` have been implemented. The core logic is derived from https://github.com/checkpoint-restore/criu/blob/criu-dev/coredump/criu_coredump/coredump.py. 

It's still a work in progress, but I would like to submit it for feedback. 

PTAL: @adrianreber, @rst0git 

Thank you!

### DIFF
--- a/crit/mempages.go
+++ b/crit/mempages.go
@@ -1,0 +1,171 @@
+package crit
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/checkpoint-restore/go-criu/v6/crit/images/mm"
+	"github.com/checkpoint-restore/go-criu/v6/crit/images/pagemap"
+)
+
+var sysPageSize = os.Getpagesize()
+
+// MemoryReader is a struct used to retrieve
+// the content of memory associated with a specific process ID (pid).
+// New instances should be created with NewMemoryReader()
+type MemoryReader struct {
+	checkpointDir  string
+	pid            uint32
+	pagesID        uint32
+	pageSize       int
+	pagemapEntries []*pagemap.PagemapEntry
+}
+
+// NewMemoryReader creates a new instance of MemoryReader with all the fields populated
+func NewMemoryReader(checkpointDir string, pid uint32, pageSize int) (*MemoryReader, error) {
+	if pageSize == 0 {
+		pageSize = sysPageSize
+	}
+
+	// Check if the given page size is a positive power of 2, otherwise return an error
+	if (pageSize & (pageSize - 1)) != 0 {
+		return nil, errors.New("page size should be a positive power of 2")
+	}
+
+	pagemapImg, err := getImg(filepath.Join(checkpointDir, fmt.Sprintf("pagemap-%d.img", pid)), &pagemap.PagemapHead{})
+	if err != nil {
+		return nil, err
+	}
+
+	pagesID := pagemapImg.Entries[0].Message.(*pagemap.PagemapHead).GetPagesId()
+
+	pagemapEntries := make([]*pagemap.PagemapEntry, 0)
+
+	for _, entry := range pagemapImg.Entries[1:] {
+		pagemapEntries = append(pagemapEntries, entry.Message.(*pagemap.PagemapEntry))
+	}
+
+	return &MemoryReader{
+		checkpointDir:  checkpointDir,
+		pid:            pid,
+		pageSize:       pageSize,
+		pagesID:        pagesID,
+		pagemapEntries: pagemapEntries,
+	}, nil
+}
+
+// GetMemPages retrieves the content of memory pages
+// associated with a given process ID (pid).
+// It retrieves the memory content within the
+// specified range defined by the start and end addresses.
+func (mr *MemoryReader) GetMemPages(start, end uint64) (*bytes.Buffer, error) {
+	size := end - start
+
+	startPage := start / uint64(mr.pageSize)
+	endPage := end / uint64(mr.pageSize)
+
+	var buffer bytes.Buffer
+
+	for pageNumber := startPage; pageNumber <= endPage; pageNumber++ {
+		var page []byte = nil
+
+		pageMem, err := mr.getPage(pageNumber)
+		if err != nil {
+			return nil, err
+		}
+
+		if pageMem != nil {
+			page = pageMem
+		} else {
+			page = bytes.Repeat([]byte("\x00"), mr.pageSize)
+		}
+
+		var nSkip, nRead uint64
+
+		if pageNumber == startPage {
+			nSkip = start - pageNumber*uint64(mr.pageSize)
+			if startPage == endPage {
+				nRead = size
+			} else {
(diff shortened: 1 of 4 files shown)


adrianreber:
Thanks for the PR. I am a bit unsure about the design. Returning the content of the pages as a return parameter does not sound right. For large memory areas I would expect that using pointers would be better and would hopefully require less data movement.

Nice to see tests included in the PR. You could try to also pass a couple of wrong parameters to test the error paths as well.

behouba:
I have made a few changes to the previous implementation:

- Added a `MemoryReader` struct to encapsulate common and reusable data.
- Modified the `GetMemPages` function to accept addresses (start and end) for flexibility, enabling the retrieval of specific portions of memory (e.g., `mm_arg`, `mm_env`, pagemap addresses etc.).
- The `GetMemPages` function now returns a *bytes.Buffer instead of []byte.

Do you think this is a  good direction for ?

crit/mempages.go
```diff
@@ -0,0 +1,134 @@
...
+	// Calculate and check the size of the memory area
+	size := end - start
+	if size == 0 {
+		return nil, fmt.Errorf("memory size must be greater than 0")
+	}
+
+	startPage := start / pageSize
+	endPage := end / pageSize
+
+	var buffer bytes.Buffer
```
adrianreber, line 58: I am not experienced enough in Go, but in C this would be wrong. You are creating a local variable and returning the pointer to it. At the end of the function the variable would be gone. Go has a garbage collector so maybe this is okay.

My expectations from C would be something like:
```go
func (mr *MemoryReader) GetMemPages(buffer *bytes.Buffer, start, end uint64) (error)  {
    // fill buffer
    return nil
}
```

Your approach seems wrong to me, but that could be just that I do not understand Go correctly.

@rst0git any comments from your side

crit/mempages.go
```diff
@@ -0,0 +1,134 @@
...
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/checkpoint-restore/go-criu/v6/crit/images/pagemap"
+)
+
+const pageSize = 4096
```
adrianreber, line 12: Please use `os.GetPageSize()` as it can be a different size.

crit/mempages_test.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+}
+
+func getTestImgsPID(dir string) (int, error) {
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return 0, fmt.Errorf("Failed to read directory: %w\n", err)
+	}
+
+	re := regexp.MustCompile(`pagemap-(\d+)\.img`)
```
adrianreber, line 114: You should be able to figure out the PID using other images. Not by parsing one of the file names.

crit/mempages.go
```diff
@@ -0,0 +1,134 @@
...
+			}
+			offset += pageSize
+		}
+
+		if !found {
+			continue
+		}
+
+		f, err := os.Open(filepath.Join(mr.checkpointDir, fmt.Sprintf("pages-%d.img", mr.pagesID)))
+		if err != nil {
```
adrianreber, line 119: Can you trigger this error cases by temporarily renaming the expected `pages-%d.img` to not exist.

crit/mempages.go
```diff
@@ -0,0 +1,134 @@
...
+			return nil, err
+		}
+
+		defer f.Close()
+
+		buff := make([]byte, pageSize)
+
+		if _, err := f.ReadAt(buff, int64(offset)); err != nil {
+			return nil, err
```
adrianreber, line 128: This error could also be triggered by supplying a too short file temporarily.

crit/mempages.go
snprajwal: The Go compiler does escape analysis during allocation, so it would identify the lifetime of the buffer pointer correctly and allocate it on the heap. You're right, this would be incorrect in a language like C, but is permissible in Go.

adrianreber: Thanks for your input. Any suggestion how to implement this. With a local variable like it is or something different?

behouba: Since both approaches will work, I am not sure which one is idiomatic Go 🤔

snprajwal: This pattern is quite common in Go, I've seen it across other projects before. I feel it's fine, the current design should work as intended. The allocator + GC will take care of things.

behouba: Agree with @snprajwal, I have seen this pattern also in other Go projects.

crit/mempages.go
behouba: Wouldn't it be better to have `pageSize` as a field of `MemoryReader` so that users of the package can set the value themselves when creating new instance of `MemoryReader` ? `os.Getpagesize()` will still be the default value.

crit/mempages.go
```diff
@@ -0,0 +1,141 @@
...
+	pagesID        uint32
+	pageSize       uint64
+	pagemapEntries []*pagemap.PagemapEntry
+}
+
+// NewMemoryReader creates a new instance of MemoryReader with all the fields populated
+func NewMemoryReader(checkpointDir string, pid uint32, pageSize uint64) (*MemoryReader, error) {
+	pSize := pageSize
+	// If the given page size is not a positive power of 2, set the default page size
+	if pageSize <= 0 || (pageSize&(pageSize-1)) != 0 {
```
rst0git, line 30: In the case where a user sets the page size to 0, it could be reasonable to fall back to a default value (i.e., sysPageSize). However, when a user specifies a page size that is not a positive power of 2, it would be more appropriate to return an error instead of silently overwriting the value.

crit/mempages.go
```diff
@@ -0,0 +1,145 @@
...
+var sysPageSize = uint64(syscall.Getpagesize())
+
+// MemoryReader is a struct used to retrieve
+// the content of memory associated with a specific process ID (pid).
+// New instances should be created with NewMemoryReader()
+type MemoryReader struct {
+	checkpointDir  string
+	pid            uint32
+	pagesID        uint32
+	pageSize       uint64
```
adrianreber, line 23: I think we should use the type we get from `Getpagesize()`. Which is just `int`. Not sure it makes sense to cast it to another value.

crit/mempages.go
```diff
@@ -0,0 +1,145 @@
...
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"syscall"
+
+	"github.com/checkpoint-restore/go-criu/v6/crit/images/pagemap"
+)
+
+var sysPageSize = uint64(syscall.Getpagesize())
```
adrianreber, line 14: Any specific reason for not using `os.Getpagesize()`. Looking at the Go sources it just calls `syscall.Getpagesize()`, but I think using `os` would be better.

rst0git (approved):
LGTM

adrianreber:
Looks very good. Thanks.

PR #134, opened 2023-06-26
feat(crit): add type to file descriptor

snprajwal:
This would be useful for us to extend the info displayed by checkpointctl. Since we're anyway holding the information with us in the object, we might as well just provide that information too.


rst0git (approved):
LGTM

PR #138, opened 2023-07-18
feat(crit): add `crit x sk` for sockets

snprajwal:
This is useful for implementing socket views in checkpointctl, going forward.

### DIFF
--- a/crit/cli/cli.go
+++ b/crit/cli/cli.go
@@ -220,9 +220,9 @@ var infoCmd = &cobra.Command{
 
 // The `crit x` command
 var xCmd = &cobra.Command{
-	Use:   "x DIR {ps|fds|mems|rss}",
+	Use:   "x DIR {ps|fd|mem|rss}",
 	Short: "Explore the image directory",
-	Long:  "Explore the image directory with one of (ps, fds, mems, rss) options",
+	Long:  "Explore the image directory with one of (ps, fd, mem, rss) options",
 	// Exactly two arguments are required:
 	// * Path of the input directory
 	// * Explore type
@@ -242,14 +242,14 @@ var xCmd = &cobra.Command{
 		switch args[1] {
 		case "ps":
 			xData, err = c.ExplorePs()
-		case "fds":
+		case "fd", "fds":
 			xData, err = c.ExploreFds()
-		case "mems":
+		case "mem", "mems":
 			xData, err = c.ExploreMems()
 		case "rss":
 			xData, err = c.ExploreRss()
 		default:
-			err = errors.New("error exploring directory: invalid explore type")
+			err = errors.New("invalid explore type (supported: {ps|fd|mem|rss})")
 		}
 		if err != nil {
 			log.Fatal(fmt.Errorf("error exploring directory: %w", err))
--- a/crit/cli/cli.go
+++ b/crit/cli/cli.go
@@ -220,9 +220,9 @@ var infoCmd = &cobra.Command{
 
 // The `crit x` command
 var xCmd = &cobra.Command{
-	Use:   "x DIR {ps|fd|mem|rss}",
+	Use:   "x DIR {ps|fd|mem|rss|sk}",
 	Short: "Explore the image directory",
-	Long:  "Explore the image directory with one of (ps, fd, mem, rss) options",
+	Long:  "Explore the image directory with one of (ps, fd, mem, rss, sk) options",
 	// Exactly two arguments are required:
 	// * Path of the input directory
 	// * Explore type
@@ -248,8 +248,10 @@ var xCmd = &cobra.Command{
 			xData, err = c.ExploreMems()
 		case "rss":
 			xData, err = c.ExploreRss()
+		case "sk":
+			xData, err = c.ExploreSk()
 		default:
-			err = errors.New("invalid explore type (supported: {ps|fd|mem|rss})")
+			err = errors.New("invalid explore type (supported: {ps|fd|mem|rss|sk})")
 		}
 		if err != nil {
 			log.Fatal(fmt.Errorf("error exploring directory: %w", err))
--- a/crit/explore.go
+++ b/crit/explore.go
@@ -71,7 +71,7 @@ func (c *crit) ExplorePs() (*PsTree, error) {
 // Fd represents the file descriptors opened in a single process
 type Fd struct {
 	PId   uint32  `json:"pId"`
-	Files []*File `json:"files,omitempty"`
+	Files []*File `json:"files"`
 }
 
 // File represents a single opened file
@@ -144,6 +144,11 @@ func (c *crit) ExploreFds() ([]*Fd, error) {
 			Fd:   "root",
 			Path: filePath,
 		})
+
+		// Omit if the process has no file descriptors
+		if len(fdEntry.Files) == 0 {
+			continue
+		}
 
 		fds = append(fds, &fdEntry)
 	}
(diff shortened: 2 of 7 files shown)


crit/cli/cli.go
```diff
@@ -242,14 +242,14 @@ var xCmd = &cobra.Command{
 		switch args[1] {
 		case "ps":
 			xData, err = c.ExplorePs()
-		case "fds":
+		case "fd":
 			xData, err = c.ExploreFds()
-		case "mems":
+		case "mem":
```
rst0git, line 247: To preserve compatibility with the Python version of `crit`, should we accept both `mems` and `mem`?

crit/Makefile
```diff
@@ -1,8 +1,10 @@
 GO ?= go
 CRIU ?= criu
+CRIT_SRC = $(shell find . -name \*.go)
```
rst0git, line 3:
```Suggestion
CRIT_SRC := $(shell find . -type f -name '*.go')
```

crit/explore.go
```diff
@@ -145,6 +145,11 @@ func (c *crit) ExploreFds() ([]*Fd, error) {
 			Path: filePath,
 		})
 
+		// Omit if the process has no file descriptors
+		if len(fdEntry.Files) == 0 {
+			continue
+		}
```
rst0git, line 151: This change is not directly related to the renaming of `fds|mems` to `fd|mem`. It should be in a separate commit.

crit/Makefile
```diff
@@ -1,8 +1,10 @@
...
+CRIT_CLI = cmd/main.go
```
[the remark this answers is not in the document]
snprajwal: I used `=` instead of `?=` since I felt it's reasonable to assume that we won't override this variable, unlike others. The CLI entrypoint will always be consistent, so I didn't see the need for `?=`.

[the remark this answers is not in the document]
rst0git: Sure, perhaps we should use [simple](https://www.gnu.org/software/make/manual/html_node/Simple-Assignment.html) (`:=`) instead of [recursive](https://www.gnu.org/software/make/manual/html_node/Recursive-Assignment.html) (`=`) assignment?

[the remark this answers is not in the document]
snprajwal: Changed and pushed in #141

crit/explore.go
```diff
@@ -373,3 +373,122 @@ func (c *crit) ExploreRss() ([]*RssMap, error) {
...
+			socket := Socket{
+				Fd:   fdInfo.GetFd(),
+				Type: fdInfo.GetType().String(),
+			}
+			switch fdInfo.GetType() {
+			case fdinfo.FdTypes_INETSK:
+				file, err := getFile(c.inputDirPath, fdInfo.GetId())
+				if err != nil {
+					return nil, err
+				}
```
rst0git, line 433: It might be good to move this code just before the `switch` statement to avoid code duplication.
```Go
file, err := getFile(c.inputDirPath, fdInfo.GetId())
if err != nil {
	return nil, err
}
```

crit/explore.go
```diff
@@ -373,3 +373,122 @@ func (c *crit) ExploreRss() ([]*RssMap, error) {
...
+// Sk represents the sockets associated with a single process
+type Sk struct {
+	PId     uint32    `json:"pId"`
+	Sockets []*Socket `json:"sockets"`
+}
+
+// Socket represents a single socket
+type Socket struct {
+	Fd       uint32 `json:"fd"`
+	Type     string `json:"type"`
```
rst0git, line 386: It might be better to use `FdType` here to avoid confusion between the fdInfo *type* and socket *type*.

crit/explore.go
```diff
@@ -373,3 +373,122 @@ func (c *crit) ExploreRss() ([]*RssMap, error) {
...
+			switch fdInfo.GetType() {
+			case fdinfo.FdTypes_INETSK:
+				file, err := getFile(c.inputDirPath, fdInfo.GetId())
+				if err != nil {
+					return nil, err
+				}
+				if isk := file.GetIsk(); isk != nil {
+					socket.State = getSkState(isk.GetState())
+					socket.Family = getSkFamily(isk.GetFamily())
+					socket.Protocol = getSkProtocol(isk.GetType())
```
rst0git, line 437: It might be better to use `socket.Type` for the value of `GetType()` instead of `socket.Protocol` to avoid confusion.

For instance, socket types are `SOCK_STREAM = 1`,  `SOCK_DGRAM = 2`, `SOCK_RAW = 3`, etc. (see `man 2 socket`)

rst0git: For example, something like the following:

```Go
...
socket.Type = getSkType(isk.GetType())
...

func getSkType(skType uint32) string {
	switch skType {
	case syscall.SOCK_STREAM:
		return "STREAM"
	case syscall.SOCK_DGRAM:
		return "DGRAM"
	case syscall.SOCK_RAW:
		return "RAW"
	case syscall.SOCK_SEQPACKET:
		return "SEQPACKET"
	...
	default:
		return "unknown"
	}
}
```

crit/utils.go
```diff
@@ -232,3 +244,97 @@ func getUnixSkFilePath(dir string, file *fdinfo.FileEntry, fID uint32) (string,
...
+		return "CLOSING"
+	case 12:
+		return "NEW_SYN_RECV"
+	default:
+		return ""
+	}
+}
+
+// Helper to identify socket family
+func getSkFamily(family uint32) string {
```
rst0git, line 301: `AF` is an abbreviation for "Address Family"
```Suggestion
func getAddressFamily(family uint32) string {
```

crit/explore.go
```diff
@@ -373,3 +373,110 @@ func (c *crit) ExploreRss() ([]*RssMap, error) {
...
+type Sk struct {
+	PId     uint32    `json:"pId"`
+	Sockets []*Socket `json:"sockets"`
+}
+
+// Socket represents a single socket
+type Socket struct {
+	Fd       uint32 `json:"fd"`
+	FdType   string `json:"fdType"`
+	Family   string `json:"family,omitempty"`
```
rst0git, line 387: Would it make sense to also add `Protocol`? For example, something like the following: 
```Go
Protocol string `json:"proto,omitempty"`
```
```Go
socket.Protocol = getProto(isk.GetProto())
```
```Go
func getProto(proto uint32) string {
	switch proto {
	case syscall.IPPROTO_ICMP:
		return "ICMP"
	case syscall.IPPROTO_ICMPV6:
		return "ICMPV6"
	case syscall.IPPROTO_IGMP:
		return "IGMP"
	case syscall.IPPROTO_RAW:
		return "RAW"
	case syscall.IPPROTO_TCP:
		return "TCP"
	case syscall.IPPROTO_UDP:
		return "UDP"
	case syscall.IPPROTO_UDPLITE:
		return "UDPLITE"
	default:
		return ""
	}
}
```

snprajwal: Ah so protocol is shortened to `proto`. I mistook that to be the descriptor for the protobuf definition :sweat_smile:

crit/utils.go
```diff
@@ -246,3 +258,129 @@ func (ps *PsTree) FindPs(pid uint32) *PsTree {
 	}
 	return nil
 }
+
+// Helper to convert slice of uint32 into IP address string
+func processIP(parts []uint32) string {
```
rst0git, line 263: Would it be possible to add unit tests for this function?

snprajwal: Yes that makes sense. I'll add it, and look into adding it for any other utility functions too if possible.

rst0git (approved):
LGTM

PR #137, opened 2023-07-11
rpc: synchronise proto bindings with the latest version of CRIU RPC p…

yevzman-swe:
rpc: synchronise proto bindings with the latest version of CRIU RPC

PR #139, opened 2023-07-20
chore(rpc): update Go bindings

snprajwal:
The rpc.proto file has been updated against criu-dev, but not the bindings. Makes #137 work, and also updates protobuf with it.


rpc/rpc.pb.go
```diff
@@ -2,8 +2,8 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.28.1
-// 	protoc        v3.19.4
+// 	protoc-gen-go v1.30.0
+// 	protoc        v4.23.4
```
rst0git, line 6: The version of protobuf is different for each distribution (Ubuntu, Fedora, Arch Linux, etc).
It might be good to update all bindings with the version we currently use in our CI.

snprajwal: The CI runners are Ubuntu 22.04, but I think the newer Ubuntu releases have a newer version of protobuf in the registry. Would we benefit by using the version available in the latest Ubuntu release registry? I'm unsure if there are any breaking changes across these versions, because the pb.go files coming in my PRs are usually through a bleeding edge protobuf installation on Arch Linux

rst0git: > Would we benefit by using the version available in the latest Ubuntu release registry?

We are using a specific version of protobuf for the criu-dev branch in CI:
https://github.com/checkpoint-restore/go-criu/blob/fe8a6277e8cc8a48dac505433d98f2f575fd741e/.github/workflows/main.yml#L53

We can update protobuf to v23.4 and all pb.go files.

rst0git (approved):
LGTM

PR #145, opened 2023-08-03
crit: add `FindPs` method on `PsTree`

behouba:
This commit introduces a new method `FindPs` on `PsTree`. This method uses a depth-first search algorithm to find a process within a process tree. This addition will be beneficial for `checkpointctl` as it allows us to easily verify the existence of a given process ID when needed.

### DIFF
--- a/crit/utils.go
+++ b/crit/utils.go
@@ -232,3 +232,17 @@ func getUnixSkFilePath(dir string, file *fdinfo.FileEntry, fID uint32) (string, 
 
 	return "unix[?]", nil
 }
+
+// FindPs performs a short-circuiting depth-first search to find
+// a process with a given PID in a process tree.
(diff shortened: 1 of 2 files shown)


crit/explore.go
```diff
@@ -24,6 +24,20 @@ type PsTree struct {
 	Children []*PsTree            `json:"children,omitempty"`
 }
 
+// FindPs performs a short-circuiting depth-first search to find
+// a process with a given PID in a process tree.
+func (c *PsTree) FindPs(pid uint32) *PsTree {
```
snprajwal, line 29:
```suggestion
func (ps *PsTree) FindPs(pid uint32) *PsTree {
```
Also, I feel this function would be better off in `utils.go`, since the functionality is not related to `crit explore`.

snprajwal (approved):
LGTM

rst0git (approved):

PR #148, opened 2023-09-05
feat: propagating ciru rpc response

nravic:
This PR is aimed to return the resp from `c.doSwrk`. It's motivated primarily by the need to access the PID of the restored process, which isn't straightforward to do without getting the result from `resp` directly.

### DIFF
--- a/main.go
+++ b/main.go
@@ -86,17 +86,17 @@ func (c *Criu) sendAndRecv(reqB []byte) ([]byte, int, error) {
 	return respB, n, nil
 }
 
-func (c *Criu) doSwrk(reqType rpc.CriuReqType, opts *rpc.CriuOpts, nfy Notify) error {
+func (c *Criu) doSwrk(reqType rpc.CriuReqType, opts *rpc.CriuOpts, nfy Notify) (*rpc.CriuResp, error) {
 	resp, err := c.doSwrkWithResp(reqType, opts, nfy, nil)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	respType := resp.GetType()
 	if respType != reqType {
-		return errors.New("unexpected CRIU RPC response")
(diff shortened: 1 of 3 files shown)


adrianreber:
The main problem is that this PR changes the API which means that everyone including go-criu has to change their code. It would be good to understand what problem this PR is trying to solve. At this point I am not in favour of this change.

nravic:
Yeah, I was afraid of the API change issue - but couldn't really see a way around it. I want access to the PID of the restored process, which I can't get without `rpc.CriuResp`. The alternative is not using `go-criu` and directly calling CRIU over RPC, as far as I can tell. 

Finding the PID of the new process is especially useful for us because we're migrating the checkpoints to new machines.

adrianreber:
Can't you use the notification mechanism? That should tell you the PID.

nravic:
https://github.com/checkpoint-restore/go-criu/blob/633daeeea34e73411f21101f0e7f6cb207153172/main.go#L166 is still private and inaccessible from outside go-criu though?

rst0git:
@nravic Could you try with something similar to [`criuNotifications`](https://github.com/opencontainers/runc/blob/319b2ba7/libcontainer/criu_linux.go#L1145) in runc?

adrianreber:
> @nravic Could you try with something similar to [`criuNotifications`](https://github.com/opencontainers/runc/blob/319b2ba7/libcontainer/criu_linux.go#L1145) in runc?

I think runc doesn't really use go-criu because it is older than go-criu. That is the kind of the main question here. Directly doing RPC calls or going through go-criu. go-criu is easier to use but abstracts away some things.

But, @nravic , if you implement the `PostRestore()` callback you should get a notification which PID the restored process has. No need to access `GetPid()`.

PR #157, opened 2023-12-20
fix(crit): use proto names in JSON

### LINKED ISSUE #154: CRIT: use proto names instead of camelCase in JSON
Currently, the CRIT package serialises protobuf to JSON with camelCase for the field names. This was chosen as a design decision to prevent having to annotate every protobuf field manually with the actual snake_case name in the marshaling/unmarshaling logic. Incidentally, I discovered that the `protojson` package now allows configuring this particular case via `protojson.MarshalOptions` ([`UseProtoNames`](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson#MarshalOptions)). Enabling this will reduce the parity between the Go and Python CRIT libraries, since the Python library uses the snake_case names from the proto definitions for the fields. Is this something we should enable? While it will not break API, it will definitely break the JSON output compared to before, and if anyone is using custom logic to parse the JSON (although the shouldn't), it will cause their package to break.

cc @rst0git @adrianreber

snprajwal:
Previously, the JSON output generated through `protojson.Marshal` was using camelCased names for the keys. It now uses snake_case, which is consistent with the behaviour of the Python implementation of CRIT.

Closes #154

cc @rst0git

### DIFF
--- a/crit/image.go
+++ b/crit/image.go
@@ -34,7 +34,7 @@ func (c *CriuEntry) MarshalJSON() ([]byte, error) {
 		return []byte(fmt.Sprint(`{"count":"`, c.Extra, `"}`)), nil
 	}
 
-	data, err := protojson.Marshal(c.Message)
+	data, err := protojson.MarshalOptions{UseProtoNames: true}.Marshal(c.Message)
 	if err != nil {
 		return nil, err
 	}


rst0git (approved):
LGTM

PR #163, opened 2024-03-06
feat(crit): add SearchPattern method on MemoryReader

behouba:
I am currently working on a feature to allow users to search for a pattern within process memory pages, as suggested by @rst0git last year. This could be used to extend `checkpointctl memparse` with search capability. 

Currently, the method `SearchPattern(pattern string) (uint64, error)` takes the pattern to search as an argument and returns the address of the first match along with an error if the pattern is not found. I am not sure about the implementation yet, and I'm considering alternative approaches:

- Instead of providing just the address of the first match, should we consider returning the entire memory pages where the pattern is found?

- Another consideration is whether to return not just the first match but all occurrences of the pattern.

@rst0git, @adrianreber , @snprajwal  PTAL.

### DIFF
--- a/crit/mempages.go
+++ b/crit/mempages.go
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 
 	"github.com/checkpoint-restore/go-criu/v7/crit/images/mm"
 	"github.com/checkpoint-restore/go-criu/v7/crit/images/pagemap"
@@ -192,4 +194,101 @@ func (mr *MemoryReader) GetShmemSize() (int64, error) {
 	}
 
 	return size, nil
+}
+
+// PatternMatch represents a match when searching for a pattern in memory.
+type PatternMatch struct {
+	Vaddr   uint64
+	Length  int
+	Context int
+	Match   string
+}
+
+// SearchPattern searches for a pattern in the process memory pages.
+func (mr *MemoryReader) SearchPattern(pattern string, escapeRegExpCharacters bool, context, chunkSize int) ([]PatternMatch, error) {
+	if context < 0 {
+		return nil, errors.New("context size cannot be negative")
+	}
+
+	// Set a default chunk size of 10MB to be read at a time
+	if chunkSize <= 0 {
+		chunkSize = 10 * 1024 * 1024
+	}
+
+	// Escape regular expression characters in the pattern
+	if escapeRegExpCharacters {
+		pattern = regexp.QuoteMeta(pattern)
+	}
+
+	regexPattern, err := regexp.Compile(pattern)
+	if err != nil {
+		return nil, err
+	}
+
+	var results []PatternMatch
+
+	f, err := os.Open(filepath.Join(mr.checkpointDir, fmt.Sprintf("pages-%d.img", mr.pagesID)))
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	for _, entry := range mr.pagemapEntries {
+		startAddr := entry.GetVaddr()
+		endAddr := startAddr + uint64(entry.GetNrPages())*uint64(mr.pageSize)
+
+		initialOffset := uint64(0)
+		for _, e := range mr.pagemapEntries {
+			if e == entry {
+				break
+			}
+			initialOffset += uint64(e.GetNrPages()) * uint64(mr.pageSize)
+		}
+
+		for offset := uint64(0); offset < endAddr-startAddr; offset += uint64(chunkSize) {
+			readSize := chunkSize
+			if endAddr-startAddr-offset < uint64(chunkSize) {
+				readSize = int(endAddr - startAddr - offset)
+			}
+
+			buff := make([]byte, readSize)
+			if _, err := f.ReadAt(buff, int64(initialOffset+offset)); err != nil {
+				if err == io.EOF {
+					break
+				}
+				return nil, err
+			}
+
+			// Replace non-printable ASCII characters in the buffer with a question mark (0x3f) to prevent unexpected behavior
+			// during regex matching. Non-printable characters might cause incorrect interpretation or premature
+			// termination of strings, leading to inaccuracies in pattern matching.
+			for i := range buff {
+				if buff[i] < 32 || buff[i] >= 127 {
+					buff[i] = 0x3F
+				}
+			}
+
+			indexes := regexPattern.FindAllIndex(buff, -1)
+			for _, index := range indexes {
+				startContext := index[0] - context
+				if startContext < 0 {
+					startContext = 0
+				}
+
+				endContext := index[1] + context
+				if endContext > len(buff) {
+					endContext = len(buff)
+				}
+
+				results = append(results, PatternMatch{
+					Vaddr:   startAddr + offset + uint64(index[0]),
+					Length:  index[1] - index[0],
+					Context: context,
+					Match:   string(buff[startContext:endContext]),
+				})
+			}
+		}
+	}
+
+	return results, nil
 }
--- a/crit/mempages_test.go
+++ b/crit/mempages_test.go
@@ -248,7 +248,7 @@ func TestGetPsArgsAndEnvVars(t *testing.T) {
 	}
 }
 
-func TestGetShmemSize(t *testing.T) {
+func TestSearchPattern(t *testing.T) {
 	pid, err := getTestImgPID()
 	if err != nil {
 		t.Fatal(err)
@@ -259,15 +259,92 @@ func TestGetShmemSize(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	size, err := mr.GetShmemSize()
-	if err != nil {
-		t.Fatal(err)
+	testCases := []struct {
+		name                   string
+		pattern                string
+		context                int
+		escapeRegExpCharacters bool
+		shouldMatch            bool
+		expectedError          error
+	}{
+		{
+			name:        "PATH environment variable",
+			pattern:     "PATH=",
+			shouldMatch: true,
+		},
+		{
+			name:        "PATH environment variable regex",
+			pattern:     `\bPATH=([^\s]+)\b`,
+			shouldMatch: true,
+		},
+		{
+			name:        "PATH environment variable regex with 10 bytes context",
+			pattern:     `\bPATH=([^\s]+)\b`,
+			context:     10,
+			shouldMatch: true,
+		},
+		{
+			name:          "PATH environment variable regex with a negative context",
+			pattern:       `\bPATH=([^\s]+)\b`,
+			context:       -1,
+			expectedError: errors.New("context size cannot be negative"),
+		},
+		{
+			name:        "PATH environment variable regex with a large context",
+			pattern:     `\bPATH=([^\s]+)\b`,
+			context:     100000,
+			shouldMatch: true,
+		},
+		{
+			name:    "Non-existent pattern",
+			pattern: "NON_EXISTENT_PATTERN",
+		},
+		{
+			name:        "PASSWORD environment variable value as regex",
+			pattern:     "123 Hello.*?",
+			shouldMatch: true,
+		},
+		{
+			name:                   "PASSWORD environment variable value with regex metacharacters to escape",
+			pattern:                `123 Hello.*?[^]@WORLD(|x)`,
+			escapeRegExpCharacters: true,
+			shouldMatch:            true,
+		},
 	}
 
-	// Verify that the shared memory size is as expected (0)
-	expectedSize := int64(0)
-	if size != expectedSize {
-		t.Fatalf("Expected shared memory size: %d, but got: %d", expectedSize, size)
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			matches, err := mr.SearchPattern(tc.pattern, tc.escapeRegExpCharacters, tc.context, 0)
+			if err != nil && tc.expectedError == nil {
+				t.Errorf("Unexpected error for pattern %s: %v", tc.pattern, err)
+			} else if err == nil && tc.expectedError != nil {
+				t.Errorf("Expected error for pattern %s: %v", tc.pattern, tc.expectedError)
+			}
+
+			if tc.shouldMatch && len(matches) == 0 {
+				t.Errorf("Expected to find a match for pattern \"%s\"", tc.pattern)
+			} else if !tc.shouldMatch && len(matches) > 0 {
+				t.Errorf("Expected not to find any match for pattern \"%s\"", tc.pattern)
+			}
+
+			for _, match := range matches {
+				content, err := mr.GetMemPages(match.Vaddr, match.Vaddr+uint64(match.Length))
+				if err != nil {
+					t.Fatalf("Failed to get memory pages: %v", err)
+				}
+
+				buff := content.Bytes()
+				for i := range buff {
+					if buff[i] < 32 || buff[i] >= 127 {
+						buff[i] = 0x3F
+					}
+				}
+
+				if !strings.Contains(match.Match, content.String()) {
+					t.Errorf("Expected to find %s in matched pattern %s", content.String(), match.Match)
+				}
+			}
+		})
 	}
 }
 
--- a/test/loop/loop.c
+++ b/test/loop/loop.c
@@ -12,6 +12,13 @@ int main(void)
 	int res = EXIT_FAILURE;
 	int start_pipe[2];
 
+	// Set a PASSWORD environment variable to test the search pattern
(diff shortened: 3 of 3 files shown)


crit/mempages.go
```diff
@@ -193,3 +194,33 @@ func (mr *MemoryReader) GetShmemSize() (int64, error) {
...
+		endAddr := startAddr + uint64(entry.GetNrPages())*uint64(mr.pageSize)
+
+		// Read the entire page content
+		pageContent, err := mr.GetMemPages(startAddr, endAddr)
+		if err != nil {
+			return 0, err
+		}
+
+		// Search for the pattern using the regular expression
+		if match := regexPattern.FindIndex(pageContent.Bytes()); match != nil {
```
snprajwal, line 218: Would it make sense to use `FindAllIndex` instead? Ideally we would want to return all matches in the memory, rather than just the first one.

behouba:
I have updated the `SearchPattern` method. Now, the function takes both the pattern and a context (specified by the number of bytes before and after the pattern). I opted for byte counts instead of line counts, considering the content of `pages-*.img` files, which doesn't neatly correspond to lines.
Also, the function now returns a slice of strings representing all occurrences of the pattern found in memory. 

cc: @adrianreber , @rst0git , @snprajwal

crit/mempages.go
```diff
@@ -193,3 +194,55 @@ func (mr *MemoryReader) GetShmemSize() (int64, error) {
...
+			if startContext < 0 {
+				startContext = startPattern
+			}
+
+			endContext := endPattern + context
+			if endContext > buffer.Len() {
+				endContext = endPattern
+			}
+
+			results = append(results, string(pagesContent[startContext:endContext]))
```
rst0git, line 243: One of the main benefits of using this functionality (in comparison to something like `grep`) is that it can display additional information about where a pattern is found in memory. Would it be possible to also include the virtual address in `results`?

behouba: Sure. Would it make sense to create a type for each result match, for example:

```
// PatternMatch represents a match when searching for a regex pattern in memory.
type PatternMatch struct {
    Vaddr        uint64       // Virtual address of the first byte of the match
    Length       int           // Length of the pattern match in bytes
    Pattern      string       // String representation of the pattern match
}
```

rst0git: Sounds good

crit/mempages.go
```diff
@@ -193,3 +194,69 @@ func (mr *MemoryReader) GetShmemSize() (int64, error) {
...
+		endAddr := startAddr + uint64(entry.GetNrPages())*uint64(mr.pageSize)
+
+		// Read the entire pages content
+		buffer, err := mr.GetMemPages(startAddr, endAddr)
+		if err != nil {
+			return nil, err
+		}
+
+		// Replace null bytes with space
+		pagesContent := bytes.ReplaceAll(buffer.Bytes(), []byte{0x0}, []byte{0x20})
```
snprajwal, line 232: This looks interesting, may I know why we're doing this? I can't seem to think of a reason to eliminate null bytes from the memory page contents.

behouba: I have found that null bytes can cause regex to either match the wrong pattern or exceed the intended matching pattern, resulting in unpredictable behavior. Replacing them with space ensures we match the intended pattern.

snprajwal (approved):

crit/mempages.go
adrianreber: You should add that information as a comment. I often here a comment should describe why you it and not what you do. So instead `Replace null bytes with space` it would be good to have the explanation why you replace the null bytes.

crit/mempages.go
```diff
@@ -193,3 +194,72 @@ func (mr *MemoryReader) GetShmemSize() (int64, error) {
...
+
+	var results []PatternMatch
+
+	for _, entry := range mr.pagemapEntries {
+		startAddr := entry.GetVaddr()
+		endAddr := startAddr + uint64(entry.GetNrPages())*uint64(mr.pageSize)
+
+		// Read the entire pages content
+		buffer, err := mr.GetMemPages(startAddr, endAddr)
+		if err != nil {
```
rst0git, line 227: @behouba What would be the performance of the search functionality if we have a container checkpoint that has pages file larger than 100 GB and checkpointctl is running on a laptop with 16 GB memory?

behouba: I guess this laptop will run out of memory because the function will try to read the entire 100GB at once. 
Thank you @rst0git for pointing this out. I will work on an optimization alternative.

rst0git: > I guess this laptop will run out of memory because the function will try to read the entire 100GB at once.

I'm not sure. It looks like [`getPage()`](https://github.com/checkpoint-restore/go-criu/blob/972bd31ea11289771b4b6cb41e80fbd2f9425d7e/crit/mempages.go#L117) reads one page at a time. However, if we have a checkpoint with 100 GB pages file (≈ 26,214,400 pages), the search functionality might be very slow.

snprajwal:
@behouba dropping a note to check if you're still working on this :)

behouba:
> @behouba dropping a note to check if you're still working on this :)

Hi @snprajwal!  Yes, I am still working on it :) 
I will soon open a draft PR on `checkpointctl` so we can assess the potential performance issues with the current implementation.

crit/mempages.go
behouba: @rst0git you were right. The previous implementation had a serious performance issue due to repeated calls to [GetMemPage()](https://github.com/checkpoint-restore/go-criu/blob/972bd31ea11289771b4b6cb41e80fbd2f9425d7e/crit/mempages.go#L69).  In my local tests with a 100MB checkpoint, the search functionality took an average of 20 seconds to complete. I optimized the implementation, and the new version is faster. Can you please take a look?

I have also opened the draft PR for checkpointctl [here](https://github.com/checkpoint-restore/checkpointctl/pull/141).

crit/mempages.go
```diff
@@ -193,3 +195,92 @@ func (mr *MemoryReader) GetShmemSize() (int64, error) {
...
+				}
+				return nil, err
+			}
+
+			// Replace non-printable ASCII characters in the buffer with spaces (0x20) to prevent unexpected behavior
+			// during regex matching. non-printable characters might cause incorrect interpretation or premature
+			// termination of strings, leading to inaccuracies in pattern matching.
+			for i := range buff {
+				if buff[i] < 32 || buff[i] >= 127 {
+					buff[i] = 0x20
```
snprajwal, line 258: Might it make more sense to replace it with a question mark `0x3f` instead of a space, so that the user can recognise the non-printable characters in the output? With a space, we effectively eliminate those characters instead - the user wouldn't know if it was originally a space or not. Of course, the ideal solution would be to use a dedicated character like the "unknown character" provided by UTF-8, but since we are processing ASCII here, a question mark would be the next best choice.

crit/mempages.go
```diff
@@ -193,3 +195,92 @@ func (mr *MemoryReader) GetShmemSize() (int64, error) {
 
 	return size, nil
 }
+
+// PatternMatch represents a match when searching for a pattern in memory.
+type PatternMatch struct {
+	Vaddr   uint64
+	Length  int
+	Context int
+	Pattern string
```
snprajwal, line 204: `Pattern` is slightly misleading here. That would be the string supplied by the user - I feel this would be better named as `Result` or `Match`.

crit/mempages.go
behouba: I agree. This would be a better approach and help the user distinguish between actual spaces and non-printable characters.

crit/mempages.go
```diff
@@ -193,3 +195,92 @@ func (mr *MemoryReader) GetShmemSize() (int64, error) {
...
+			buff := make([]byte, readSize)
+			if _, err := f.ReadAt(buff, int64(initialOffset+offset)); err != nil {
+				if err == io.EOF {
+					break
+				}
+				return nil, err
+			}
+
+			// Replace non-printable ASCII characters in the buffer with question mark (0x3f) to prevent unexpected behavior
+			// during regex matching. non-printable characters might cause incorrect interpretation or premature
```
snprajwal, line 254:
```suggestion
			// during regex matching. Non-printable characters might cause incorrect interpretation or premature
```

crit/mempages.go
```diff
@@ -193,3 +195,92 @@ func (mr *MemoryReader) GetShmemSize() (int64, error) {
...
+		return nil, errors.New("context size cannot be negative")
+	}
+
+	regexPattern, err := regexp.Compile(pattern)
+	if err != nil {
+		return nil, err
+	}
+
+	var results []PatternMatch
+	chunkSize := 10 * 1024 * 1024 // Set chunk size of 10MB to be read at a time
```
snprajwal, line 219: I was wondering if you've tried to play around with this number and see how the read efficiency is affected. For very large numbers there will be a significant performance dip, but I'm curious if there is an optimum chunk size based on the hardware that the program runs on. Should we allow this to be customisable?

behouba: On my setup (6-core AMD Ryzen 5 5600U with 16GB of RAM), there wasn't any noticeable improvement in read efficiency when running benchmarks with different chunk sizes on a 10GB checkpoint archive. It might be beneficial to make this value customizable, allowing users to adjust the chunk size based on their specific hardware and needs.

Here are the results from one of my benchmark. Repeated runs of the same benchmark showed no significant performance differences.

```
Chunk Size: 1 MB, Time Taken: 18.983292ms
+------------------+------------------------------------+----------+
|     ADDRESS      |               MATCH                | INSTANCE |
+------------------+------------------------------------+----------+
| 000000c000014020 | ontainer=podmanHOME=/root????????? |        1 |
+------------------+------------------------------------+----------+
| 00007ffcd31e5fcb | ompute,utility?HOME=/root?HOSTNAME |        2 |
+------------------+------------------------------------+----------+

Chunk Size: 5 MB, Time Taken: 20.797069ms
+------------------+------------------------------------+----------+
|     ADDRESS      |               MATCH                | INSTANCE |
+------------------+------------------------------------+----------+
| 000000c000014020 | ontainer=podmanHOME=/root????????? |        1 |
+------------------+------------------------------------+----------+
| 00007ffcd31e5fcb | ompute,utility?HOME=/root?HOSTNAME |        2 |
+------------------+------------------------------------+----------+

Chunk Size: 10 MB, Time Taken: 19.223068ms
+------------------+------------------------------------+----------+
|     ADDRESS      |               MATCH                | INSTANCE |
+------------------+------------------------------------+----------+
| 000000c000014020 | ontainer=podmanHOME=/root????????? |        1 |
+------------------+------------------------------------+----------+
| 00007ffcd31e5fcb | ompute,utility?HOME=/root?HOSTNAME |        2 |
+------------------+------------------------------------+----------+

Chunk Size: 50 MB, Time Taken: 19.454568ms
+------------------+------------------------------------+----------+
|     ADDRESS      |               MATCH                | INSTANCE |
+------------------+------------------------------------+----------+
| 000000c000014020 | ontainer=podmanHOME=/root????????? |        1 |
+------------------+------------------------------------+----------+
| 00007ffcd31e5fcb | ompute,utility?HOME=/root?HOSTNAME |        2 |
+------------------+------------------------------------+----------+

Chunk Size: 100 MB, Time Taken: 19.297328ms
+------------------+------------------------------------+----------+
|     ADDRESS      |               MATCH                | INSTANCE |
+------------------+------------------------------------+----------+
| 000000c000014020 | ontainer=podmanHOME=/root????????? |        1 |
+------------------+------------------------------------+----------+
| 00007ffcd31e5fcb | ompute,utility?HOME=/root?HOSTNAME |        2 |
+------------------+------------------------------------+----------+

Chunk Size: 500 MB, Time Taken: 18.524651ms
+------------------+------------------------------------+----------+
|     ADDRESS      |               MATCH                | INSTANCE |
+------------------+------------------------------------+----------+
| 000000c000014020 | ontainer=podmanHOME=/root????????? |        1 |
+------------------+------------------------------------+----------+
| 00007ffcd31e5fcb | ompute,utility?HOME=/root?HOSTNAME |        2 |
+------------------+------------------------------------+----------+

Chunk Size: 1000 MB, Time Taken: 17.56645ms
+------------------+------------------------------------+----------+
|     ADDRESS      |               MATCH                | INSTANCE |
+------------------+------------------------------------+----------+
| 000000c000014020 | ontainer=podmanHOME=/root????????? |        1 |
+------------------+------------------------------------+----------+
| 00007ffcd31e5fcb | ompute,utility?HOME=/root?HOSTNAME |        2 |
+------------------+------------------------------------+----------+

Chunk Size: 2000 MB, Time Taken: 19.011025ms
+------------------+------------------------------------+----------+
|     ADDRESS      |               MATCH                | INSTANCE |
+------------------+------------------------------------+----------+
| 000000c000014020 | ontainer=podmanHOME=/root????????? |        1 |
+------------------+------------------------------------+----------+
| 00007ffcd31e5fcb | ompute,utility?HOME=/root?HOSTNAME |        2 |
+------------------+------------------------------------+----------+
```

snprajwal: ~20s for a 10GB checkpoint is quite good! Thanks for the improvements :)

snprajwal (approved):
LGTM

rst0git:
@behouba Thank you for working on this feature and updating the pull request!
Would it be possible to apply the recent changes in the [PR for checkpointctl](https://github.com/checkpoint-restore/checkpointctl/pull/141) before we merge this pull request?

crit/mempages.go
```diff
@@ -193,3 +195,96 @@ func (mr *MemoryReader) GetShmemSize() (int64, error) {
...
+	}
+
+	regexPattern, err := regexp.Compile(pattern)
+	if err != nil {
+		return nil, err
+	}
+
+	var results []PatternMatch
+
+	f, err := os.Open(filepath.Join(mr.checkpointDir, fmt.Sprintf("pages-%d.img", mr.pagesID)))
```
rst0git, line 225: When testing the changes in the current PR for checkpointctl (with a checkpoint crated with Podman), the search functionality fails with the following error:
```
$ sudo ./checkpointctl memparse -s test /tmp/chkpt.tar.gz 
Error: failed to create memory reader: error opening binary file: open /tmp/checkpointctl4159436349/checkpoint/pagemap-0.img: no such file or directory

$ sudo tar tvf /tmp/chkpt.tar.gz | grep pagemap
-rw-r--r-- root/root       254 2024-07-25 17:24 checkpoint/pagemap-1.img
```

behouba: Oh, I see where this comes from.  This is due to the fact that the search is done per PID, so a PID need to be provided. I will add a more meaningful error message, or should we allow the search to go through all PIDs?

behouba: > Should we check that the value of `mr.pagesID` is >= 1 here?

Yes, I will add a check for valid PID.

rst0git: > should we allow the search to go through all PIDs?

It would make sense to support search through all PIDs with something like `-a, --all` option.

snprajwal: I would suggest that for now, we can check if the PID > 0 and allow searching only for a single PID. We can add the feature to search across all PIDs in a separate PR.

behouba: Wouldn't it be better check that PID is valid as we did [here](https://github.com/checkpoint-restore/checkpointctl/blob/723e647f0a1fe188ee6a3953570843aa546802a5/cmd/memparse.go#L172)  so that we can output a more meaningful error message? Checking if PID > 0 won't guarantee that the PID is valid, and the user will receive the same error message otherwise.

rst0git: > Searching through all PIDs would be a little more involved since we'd have to query the process tree and fetch all PIDs from there

Perhaps it would be better to keep the functionality in go-criu simple and use something like a loop in checkpointctl that iterates over all PIDs.

crit/mempages.go
```diff
@@ -193,3 +195,96 @@ func (mr *MemoryReader) GetShmemSize() (int64, error) {
...
+	if context < 0 {
+		return nil, errors.New("context size cannot be negative")
+	}
+
+	// Set a default chunk size of 10MB to be read at a time
+	if chunkSize <= 0 {
+		chunkSize = 10 * 1024 * 1024
+	}
+
+	regexPattern, err := regexp.Compile(pattern)
```
rst0git, line 218: It is very common for passwords to contain characters like `*` `|` `.` `(` `)`. However, these are regex metacharacters that users would need to manually escape when searching.

Would it make sense to use a boolean argument indicating whether to escapes metacharacters?

For example, something like the following:

```Go
if escapeRegExpCharacters {
    pattern = regexp.QuoteMeta(pattern)
}
```
with `checkpointctl memparse` providing two options, where the first option escapes metacharacters:
- `--search=<string pattern>`
- `--search-regex=<regex pattern>`

behouba: This completely makes sense. Thanks @rst0git )

rst0git (approved):
LGTM

PR #165, opened 2024-03-12
Add functions which are used in Podman, CRI-O and containerd

adrianreber:
Convenience functions to check if CRIU is available and at least a certain version are now used in containerd, CRI-O and Podman. This change moves those functions to go-criu.

I am not entirely sure about the naming and the location. If this used somewhere else it would be:
```
import "github.com/checkpoint-restore/go-criu/v7/utils"
...
utils.GetCriuVersion()
```
Which results in `utils`, which is a very generic name. Maybe changing the directory name to `crutils` would be better?

```
import "github.com/checkpoint-restore/go-criu/v7/crutils"
...
crutils.GetCriuVersion()
```

Not sure. Any ideas about naming to make it easy to include and understand where it is coming from?

### DIFF
--- a/utils/criu_linux.go
+++ b/utils/criu_linux.go
@@ -1,0 +1,42 @@
+package utils
+
+import (
+	"fmt"
+
+	"github.com/checkpoint-restore/go-criu/v7"
+	"github.com/checkpoint-restore/go-criu/v7/rpc"
+	"google.golang.org/protobuf/proto"
+)
+
+// CheckForCRIU checks if CRIU is available and if it is as least the
+// version as specified in the "version" parameter.
+func CheckForCriu(version int) error {
+	criuVersion, err := GetCriuVersion()
+	if err != nil {
+		return fmt.Errorf("failed to check for criu version: %w", err)
+	}
+
+	if criuVersion >= version {
+		return nil
+	}
+	return fmt.Errorf("checkpoint/restore requires at least CRIU %d, current version is %d", version, criuVersion)
+}
+
+// Convenience function to easily check if memory tracking is supported.
+func IsMemTrack() bool {
+	features, err := criu.MakeCriu().FeatureCheck(
+		&rpc.CriuFeatures{
+			MemTrack: proto.Bool(true),
+		},
+	)
+	if err != nil {
+		return false
+	}
+
+	return features.GetMemTrack()
+}
+
+func GetCriuVersion() (int, error) {
+	c := criu.MakeCriu()
+	return c.GetCriuVersion()
+}
--- a/test/main.go
+++ b/test/main.go
@@ -3,11 +3,13 @@ package main
 import (
 	"fmt"
 	"log"
+	"math"
 	"os"
 	"strconv"
 
 	"github.com/checkpoint-restore/go-criu/v7"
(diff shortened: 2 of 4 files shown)


rst0git:
@adrianreber It is common to use import aliases in Go. For example, the alias [`criu`](https://github.com/opencontainers/runc/blob/109a7a047823426e7748abf347b47d2500570d86/libcontainer/criu_opts_linux.go#L3
) is used in runc:
```Go
import criu "github.com/checkpoint-restore/go-criu/v6/rpc"
```

utils/criu_linux.go
```diff
@@ -0,0 +1,45 @@
+//go:build linux
+// +build linux
```
snprajwal, line 2: Since the filename is suffixed with `_linux`, it will be included only in Linux builds. Do you think we need to keep the build constraints within the file too? It seems redundant to me.

utils/criu_linux.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	if criuVersion >= version {
+		return nil
+	}
+	return fmt.Errorf("checkpoint/restore requires at least CRIU %d, current version is %d", version, criuVersion)
+}
+
+// Convenience function to easily check if memory tracking is supported.
+func MemTrack() bool {
```
snprajwal, line 29: Would it make more sense to name this `IsMemTrack()`?

adrianreber:
> @adrianreber It is common to use import aliases in Go. For example, the alias [`criu`](https://github.com/opencontainers/runc/blob/109a7a047823426e7748abf347b47d2500570d86/libcontainer/criu_opts_linux.go#L3) is used in runc:
> 
> ```go
> import criu "github.com/checkpoint-restore/go-criu/v6/rpc"
> ```

Right, but the question is, if we come with a good name, then maybe the alias would not be necessary.

utils/criu_linux.go
adrianreber: Ah, good question. Just recently I had a problem with something like that. Let me try it again.

utils/criu_linux.go
adrianreber: Sure. Will do.

adrianreber: I changed it to `GetMemTrack()`. But I am also open to `IsMemTrack()`. Both approaches seem valid. I used `Get..` as that is used in one of the functions.

utils/criu_linux.go
snprajwal: Yes, that would still need the constraints, but we can omit it just for this file.

utils/criu_linux.go
snprajwal: The way I see it, `Get` signifies the actual object - in this case, possibly a handler or a struct that can work with memory tracking. `Is` signifies a purely existential check, which seems more appropriate in this context.

adrianreber: Changed to `IsMemTrack()`

adrianreber:
The missing test coverage seems to mainly in the test code. Not sure it makes sense to track test coverage in the actual test case :wink:

snprajwal (approved):

rst0git (approved):
LGTM

PR #166, opened 2024-03-14
fix(coverage): ignore test files

snprajwal:
Codecov ignores unit tests by default, but does not recognise the `test` directory for integration tests. Tests are written to generate coverage, so it does not make sense to check the coverage of the tests themselves.


adrianreber (approved):
Good idea. Thanks!

PR #175, opened 2024-06-27
criu swrk: return child error to caller

Luap99:
In the podman CI we are seeing a weird flake during criu version detection[1]. The write to the socket just fails with broken pipe. The logical thing to assume here is that the child exited. However the current code never reports back the child error from wait nor does it try to capture the output from it. This fixes both. The cleanup error is now added to the returned error so the caller sees both.

As errors.Join is used from the std lib bump the minimum go version to 1.20.

[1] https://github.com/containers/podman/issues/18856

### DIFF
--- a/main.go
+++ b/main.go
@@ -61,13 +61,19 @@ func (c *Criu) Prepare() error {
 }
 
 // Cleanup cleans up
-func (c *Criu) Cleanup() {
+func (c *Criu) Cleanup() error {
+	var errs []error
 	if c.swrkCmd != nil {
-		c.swrkSk.Close()
+		if err := c.swrkSk.Close(); err != nil {
+			errs = append(errs, err)
+		}
 		c.swrkSk = nil
-		_ = c.swrkCmd.Wait()
+		if err := c.swrkCmd.Wait(); err != nil {
+			errs = append(errs, fmt.Errorf("criu swrk failed: %w", err))
+		}
 		c.swrkCmd = nil
 	}
+	return errors.Join(errs...)
 }
 
 func (c *Criu) sendAndRecv(reqB []byte) ([]byte, int, error) {
@@ -99,9 +105,7 @@ func (c *Criu) doSwrk(reqType rpc.CriuReqType, opts *rpc.CriuOpts, nfy Notify) e
 	return nil
 }
 
-func (c *Criu) doSwrkWithResp(reqType rpc.CriuReqType, opts *rpc.CriuOpts, nfy Notify, features *rpc.CriuFeatures) (*rpc.CriuResp, error) {
-	var resp *rpc.CriuResp
-
+func (c *Criu) doSwrkWithResp(reqType rpc.CriuReqType, opts *rpc.CriuOpts, nfy Notify, features *rpc.CriuFeatures) (resp *rpc.CriuResp, retErr error) {
 	req := rpc.CriuReq{
 		Type: &reqType,
 		Opts: opts,
@@ -121,7 +125,13 @@ func (c *Criu) doSwrkWithResp(reqType rpc.CriuReqType, opts *rpc.CriuOpts, nfy N
 			return nil, err
 		}
 
-		defer c.Cleanup()
+		defer func() {
+			// append any cleanup errors to the returned error
+			err := c.Cleanup()
+			if err != nil {
+				retErr = errors.Join(retErr, err)
+			}
+		}()
 	}
 
 	for {
--- a/phaul/client.go
+++ b/phaul/client.go
@@ -1,6 +1,7 @@
 package phaul
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/checkpoint-restore/go-criu/v7"
@@ -55,7 +56,7 @@ func isLastIter(iter int, stats *stats.DumpStatsEntry, prevStats *stats.DumpStat
 }
 
 // Migrate function
-func (pc *Client) Migrate() error {
+func (pc *Client) Migrate() (retErr error) {
 	criu := criu.MakeCriu()
 	psi := rpc.CriuPageServerInfo{
 		Fd: proto.Int32(int32(pc.cfg.Memfd)),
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/checkpoint-restore/go-criu/v7
 
-go 1.18
+go 1.20
```
rst0git, line 3: @Luap99 Would you be able to also update the CI workflow in [main.yml](https://github.com/checkpoint-restore/go-criu/blob/ae18f4225e2e1107609640a7d50242eaa4b41536/.github/workflows/main.yml#L16) with this change?

main.go
```diff
@@ -44,9 +47,12 @@ func (c *Criu) Prepare() error {
 	srv := os.NewFile(uintptr(fds[1]), "criu-xprt-srv")
 	defer srv.Close()
 
+	out := new(bytes.Buffer)
 	args := []string{"swrk", strconv.Itoa(fds[1])}
 	// #nosec G204
 	cmd := exec.Command(c.swrkPath, args...)
+	cmd.Stdout = out
+	cmd.Stderr = out
```
snprajwal, line 55: Would it make more sense to capture stdout and stderr separately? I'd assume you wouldn't want random info logs to clutter the error logs while debugging

main.go
```diff
@@ -99,7 +112,7 @@ func (c *Criu) doSwrk(reqType rpc.CriuReqType, opts *rpc.CriuOpts, nfy Notify) e
 	return nil
 }
 
-func (c *Criu) doSwrkWithResp(reqType rpc.CriuReqType, opts *rpc.CriuOpts, nfy Notify, features *rpc.CriuFeatures) (*rpc.CriuResp, error) {
+func (c *Criu) doSwrkWithResp(reqType rpc.CriuReqType, opts *rpc.CriuOpts, nfy Notify, features *rpc.CriuFeatures) (_ *rpc.CriuResp, retErr error) {
```
snprajwal, line 115: It might be better to name the first return value as `resp`, rather than leaving it unnamed.

main.go
rst0git: @snprajwal I believe that CRIU uses only one file descriptor for logging ([`log_get_fd()`](https://github.com/checkpoint-restore/criu/blob/bf8c134714d6e041261b5129561eeb49aa734ec1/criu/log.c#L79)).

Luap99: yeah I wasn't sure to be honest, I did a quick look in the criu source and found perror calls there that I might be interested i. But I wasn't sure about stdout, if it doesn't log there it should not make a difference but I could also drop it back to /dev/null like it is right now.

go.mod
Luap99: sure

Luap99:
Ok so I understand why the CI tests hang.

When capturing output like this in go, the go std lib creates a pipe that is setup on stout/err in the child. After wait() is called on the child go then tries to read all remaining output from the pipe fd set one stdout/err on the child until EOF. However this has one major problem, when the stdout/err file descriptors in the child are then given to other processes. In the hanging tests case this seems to be in the case of CriuReqType_PAGE_SERVER_CHLD which causes a normal fork in the criu swrk process to create criu: page-server which inherits stdout/err.

https://github.com/checkpoint-restore/criu/blob/bf8c134714d6e041261b5129561eeb49aa734ec1/criu/cr-service.c#L1013

I don't have a good answer for this, either we try to make criu close the std fds (dups them to /dev/null) when it forks that there or we do not capture the output at all. Or maybe we special case CriuReqType_PAGE_SERVER_CHLD to not capture output in only this case but I have no idea about criu at all so I don't know if this would effect more options.

Of course I would like to get the output because without debugging failures is basically impossible.

Luap99:
@rst0git There is no point in rebasing this. I need an answer on the issue raised above. So either I add a special exclude CriuReqType_PAGE_SERVER_CHLD (not sure if there are more we need to worry about) from output capture or I do not try to capture the output at all. I like the output capture of course.

rst0git:
@Luap99 In general, CRIU writes all output to a log file (e.g., https://github.com/opencontainers/runc/commit/c77aaa3f95ec7cffbb24c53b0ee8a0b6d19d8839). We don't capture any output from stdout for `criu swrk` (e.g., [libcriu](https://github.com/checkpoint-restore/criu/blob/5c3f6217f44acf1edec57a8abf4edcb624accd6f/lib/c/criu.c#L1397), [pycriu](https://github.com/checkpoint-restore/criu/blob/5c3f6217f44acf1edec57a8abf4edcb624accd6f/lib/pycriu/criu.py#L106)), except in [runc](https://github.com/opencontainers/runc/blob/3778ae603c706494fd1e2c2faf83b406e38d687d/libcontainer/criu_linux.go#L927), where stdin, stdout, and stderr are set to [inherit stdio](https://github.com/opencontainers/runc/blob/3778ae603c706494fd1e2c2faf83b406e38d687d/tty.go#L66).

Luap99:
I am not talking about any container runtime involvement here. The fact is that this library just fails with a useless error (write criu-xprt-cln: broken pipe) when the swrk process fails. I can capture the wait error but than we only get the exit code not whatever errors criu pritned to stderr, i.e. 
https://github.com/checkpoint-restore/criu/blob/5c3f6217f44acf1edec57a8abf4edcb624accd6f/lib/c/criu.c#L1398

Are you telling me this library should always create a log when starting the swrk process? And then read this file after wait?

rst0git:
@avagin Do you have any ideas on how we could capture the CRIU logs in go-criu?

For example, the following repository contains a simple program that prints the version of CRIU:
https://github.com/rst0git/test-go-criu-swrk
```
$ go build
$ ./test-go-criu-swrk 
2024/07/19 14:51:09 31900
```

If we create the following configuration file, it will cause `criu swrk` to always fail with an error:
```
$ echo  test | sudo tee /etc/criu/default.conf
$ criu swrk 2
criu: unrecognized option '--test'
...
```
However, this error is not captured in stdout/stderr: 
```
$ ./test-go-criu-swrk 
2024/07/19 14:49:53 read criu-xprt-cln: connection reset by peer
criu swrk failed: exit status 1 (<nil>)
```

A similar problem appears with the CI tests for Podman (https://github.com/containers/podman/issues/18856).

Luap99:
Ok I understand now, I got confused by the perror lines then as I assumed they print to stderr but if they redirect it into the a log file then it will not help us.

Luap99:
I guess I push my change without the output capture for now, getting a exit status 1 or killed would already tell us more useful things in podman.

rst0git (approved):
LGTM

PR #177, opened 2024-06-27
chore(magicgen): migrate script to Python

### LINKED ISSUE #176: Migrate magic-gen script to Python
The [magic-gen script](https://github.com/checkpoint-restore/go-criu/blob/master/scripts/magic-gen/magicgen.go) is currently written in Go. While it works, it would probably be much simpler to have this in Python, just to make it neater, and also be consistent with other generation scripts (gen-proto is in Python).

snprajwal:
The old script was in Go, and naively split the lines by whitespace to identify the magic names and values. The new Python script uses a proper regex rule to parse the input and generate the output.

Fixes #176


magic/magic.go
```diff
@@ -30,8 +30,6 @@ func LoadMagic() MagicMap {
 	magicMap.ByValue[1631600659] = "CPUINFO"
 	magicMap.ByName["CREDS"] = 1409430855
 	magicMap.ByValue[1409430855] = "CREDS"
-	magicMap.ByName["CRTOOLS_IMAGES_V1_1"] = 2
-	magicMap.ByValue[2] = "CRTOOLS_IMAGES_V1_1"
```
rst0git, line 34: Why are these two lines being removed?

snprajwal: Ah, the regex parsing does not match non-hex values, so these lines are left out from the new run. Would we need these though? They are not image types themselves, so I'd assume they're not directly used by go-criu?

rst0git: [`CRTOOLS_IMAGES_V1_1`](https://github.com/checkpoint-restore/criu/blob/c2f9f900a94cbfefda9385f7dd1b757b226bbaa2/criu/include/magic.h#L13) is used as a common magic for protobuf all [images](https://criu.org/Images) except inventory and service files:
```
IMAGE_FILE ::= MAGIC [MAGIC_2] { ENTRY }
```

See https://github.com/checkpoint-restore/criu/commit/f5ea330ce1f38486a9f80d68504899d7dad48cb3

scripts/magic-gen/magicgen.py
```diff
@@ -0,0 +1,77 @@
...
+
+    try:
+        with open(args.dest, 'w') as dest:
+            write_magics(dest, magics)
+    except IOError:
+        sys.exit("Failed to open magic.py")
+
+def read_magics(src):
+    magics = {}
+    pattern = re.compile(r"#define\s+(\S+)\s+0x([0-9A-Fa-f]+)")
```
rst0git, line 29: Could you please add a comment to describe the patterns matched by this regular expression? This will be particularly useful for new contributors who are not familiar with the magic-gen scripts.

scripts/magic-gen/Makefile
```diff
@@ -1,17 +1,14 @@
...
 magic-gen: clean magic.h
-	$(GO) run magicgen.go magic.h $(MAGIC_DEST)
+	@$(PY) magicgen.py magic.h $(MAGIC_DEST)
 
 magic.h:
 	curl -s https://raw.githubusercontent.com/checkpoint-restore/criu/criu-dev/criu/include/magic.h -o magic.h
 
 test:
-	@echo "Running unit test"
-	$(GO) test -v
```
rst0git, line 13: Would it be possible to add unit tests for `read_magics` / `write_magics` similar to `magicgen_test.go`?

rst0git (approved):
Well done!

PR #185, opened 2024-07-19
ci: moving back to tokenless

rst0git:


snprajwal:
Are we sure this is a good idea? We definitely want to be tracking coverage, and doing this change would mean that if the token gets removed from the repo in the future, the pipeline will still erroneously succeed. Would it not be better to just let it fail if the token is not set?

adrianreber:
Silently skipping something might indeed result that we forget about this again. What was your motivation for this change @rst0git. Maybe that will help us to understand this PR better.

rst0git:
> Would it not be better to just let it fail if the token is not set?

@snprajwal Adrian has configured the value for `CODECOV_TOKEN` in [repository secrets](https://docs.codecov.com/docs/adding-the-codecov-token#github-actions). However, the GitHub action is unable to use it. This problem does not occur when running the CI workflow in my fork configured with codecov token for my personal repository.

```
##[debug]..Evaluating secrets:
##[debug]..=> Object
##[debug]..Evaluating String:
##[debug]..=> 'CODECOV_TOKEN'
##[debug]=> null
```
We can see the code coverage in the CI logs but uploading without a token is not supported:
```
# Print coverage from this run
go tool covdata percent -i=/home/runner/work/go-criu/go-criu/test/.coverage
	command-line-arguments		coverage: 68.4% of statements
...
```

> What was your motivation for this change

This is a workaround until we confirm that the codecov configuration works or figure out how to fix the problem.

snprajwal:
@adrianreber do you want to double-check if the secret is set correctly please? It could be an accidental typo that's stopping the workflow from recognising it

snprajwal:
Dropping this here too - the coverage workflow on checkpointctl works fine: https://github.com/checkpoint-restore/checkpointctl/actions/runs/10009967467/job/27670052231

PR #186, opened 2024-08-28
release: v7.2.0

snprajwal:
cc @behouba

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,8 @@ The following table shows the relation between go-criu and criu versions:
 
 | Major version  | Latest release | CRIU version |
 | -------------- | -------------- | ------------ |
-| v7             | 7.1.0          | 3.18         |
+| v7             | 7.2.0          | 3.19         |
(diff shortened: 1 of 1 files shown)


snprajwal:
Okay, looks like we'll need some fixes before making the release. I'm away until Sunday, I'll do it once I'm back.

snprajwal:
@rst0git it's interesting how I was able to create this branch through the web UI, but now it's protected - I can neither push to it, nor delete it. I meant to create the branch on my fork, but ended up creating it on this repo instead. Do we apply protection rules on non-default branches too?

rst0git:
> Do we apply protection rules on non-default branches too?

I am not sure, we usually don't create branches in the main repository. It might be better to open another pull request and we can close this one.

@adrianreber do you have permissions to delete this branch?

adrianreber:
Branch protection is on all branches (`*`). I did not set up the branch protection, but it does not sound bad. I removed the `release` branch now.

PR #188, opened 2024-09-01
ci(golangci-lint): disable gosec G115

snprajwal:
Changes were made to this check in v1.60.2, which introduced a lot of noise and false positives. This commit disables this for now.

Ref: https://github.com/securego/gosec/issues/1185


snprajwal:
ping @rst0git

adrianreber (approved):
Makes sense.

PR #208, opened 2025-03-28
ci: bump actions, remove Go matrix

snprajwal:
The Go matrix is unnecessary, as Go provides backwards compatibility across existing APIs, and we are almost always just one version behind the latest one.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,6 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        go-version: [1.21.x, 1.22.x, 1.23.x]
         criu-branch: [master, criu-dev]
         exclude: # Skip duplicate CRIT build tests on macos/windows
           - os: windows-latest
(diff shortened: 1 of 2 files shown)


rst0git:
> Go matrix is unnecessary, as Go provides backwards compatibility

@snprajwal I believe the purpose of the Go matrix is to check that building go-criu works with older versions of Go. For example, if we use a feature that is available only in 1.24, it could break users that require previous versions of Go.

snprajwal:
> @snprajwal I believe the purpose of the Go matrix is to check that building go-criu works with older versions of Go. For example, if we use a feature that is available only with >=1.24, it could break users that require previous versions of Go.

I understand that, but our `go.mod` specifies 1.23, which means that any library user would also be required to have a minimum Go version of 1.23 for their module. I think we would not be responsible if the user uses the library with a lower Go version - it can also break due to a transitive dependency of go-criu that does not support their version.

adrianreber:
We definitely do not need to test anything older than what is in `go.mod` specified as minimum.

adrianreber:
This PR works for me.

PR #210, opened 2025-05-30
Invalidate global vars containing parsed images in Explore* methods

pbeschetnov:
Caching common criu images in global variables doesn't work in these scenarios:
- crit.Explore...() are used in different img directories
- between crit.Explore...() img files are updated (new dump happened)

With this PR I'm fixing this problem and supporting listed cases by invalidating the caches during each Explore call.

### DIFF
--- a/crit/explore.go
+++ b/crit/explore.go
@@ -26,6 +26,8 @@ type PsTree struct {
 
 // ExplorePs constructs the process tree and returns the root process
 func (c *crit) ExplorePs() (*PsTree, error) {
+	invalidateCache()
+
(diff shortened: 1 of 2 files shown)


snprajwal:
It looks like you were already doing this in the first commit :) that would be nice to have. Please note that we use self-contained commits without follow-up fixup commits -please squash your commits into a single one. It would also be great if you could write a descriptive commit message that explains the intent of the commit better. Thanks!

rst0git (approved):
LGTM

@pbeschetnov Thank you for fixing this!

PR #246, opened 2026-06-04
rpc, stats: switch to protobuf-go-lite

kolyshkin:
This is mostly done to desperately save runc (and podman) binary sizes, of which google protobuf is a hefty part.

----

Regenerate the top-level rpc/ and stats/ packages with protoc-gen-go-lite instead of protoc-gen-go, so the embeddable library path (the root, phaul and utils packages used by downstream consumers) no longer pulls in the reflection-based google.golang.org/protobuf runtime.

Lite messages expose MarshalVT/UnmarshalVT instead of the generic proto.Marshal/Unmarshal, and lite does not ship the proto2 scalar constructors (String/Int32/Bool); these are reimplemented in a small internal/proto package, imported aliased as proto so call sites are unchanged.

crit/ and crit/images/* (including the separate crit/images/stats package) stay on google.golang.org/protobuf, as crit's generic decode and JSON handling rely on reflection (proto.Clone, protojson, proto.Message in its public Critter interface).


kolyshkin:
For runc this is being tested in https://github.com/opencontainers/runc/pull/5312. Will open a podman test PR soon.

kolyshkin:
Alas podman is using protobuf in a few other places so there's no (easy) way to eliminate it entirely. For runc though it's a clear win.

kolyshkin:
This is ready for review, any feedback is welcome.

snprajwal (approved):
As per https://github.com/opencontainers/runc/pull/5312#issuecomment-4624854385, this shaves off ~2MB for downstream consumers of the library. That's a big win!

The changes LGTM, as long as this third-party protobuf library continues to be actively maintained.

adrianreber:
Thanks. This sounds like a good idea. But probably means that we will to have to make a v9, right? This sounds like it could count as breaking API.

rst0git (commented):
LGTM

kolyshkin:
> Thanks. This sounds like a good idea. But probably means that we will to have to make a v9, right? This sounds like it could count as breaking API.

Do you want bumping to v9 in this PR?

adrianreber:
No. I think the v9 bump can happen separately.

snprajwal:
I'm not sure if this requires a v9 bump. I don't see any public APIs or behaviour changing; only the implementation details have changed, which are opaque to library consumers.

We are introducing new public API, but AFAIK that can be included in a minor release since it's non-breaking.

adrianreber:
> I'm not sure if this requires a v9 bump. I don't see any public APIs or behaviour changing; only the implementation details have changed, which are opaque to library consumers.
> 
> We are introducing new public API, but AFAIK that can be included in a minor release since it's non-breaking.

That is correct, but pulling partially completely different dependencies might be unexpected to users. Not sure.

kolyshkin:
Added the second commit which uses `tool` in go.mod to pin the needed tools to specific versions, and use go install tool to install those from CI, slightly simplifying setup and making CI runs more predictable. PTAL

snprajwal:
Nice one, thanks for that update. You'd need to update that commit with a signoff to make CI happy though.

rst0git (approved):

rst0git:
> I think the v9 bump can happen separately.

@adrianreber Would you like to create a release for v9?

adrianreber:
> > I think the v9 bump can happen separately.
> 
> @adrianreber Would you like to create a release for v9?

Do we still think a v9 is necessary? @snprajwal's comment that the API does not change sounded convincing to me.

kolyshkin:
> > @adrianreber Would you like to create a release for v9?
> 
> Do we still think a v9 is necessary? @snprajwal's comment that the API does not change sounded convincing to me.

I'm fine with v8.3.0 or so -- API is still the same, the underlying implementation changed and a new dependency is needed. I think that as long as release notes explain what is done we're fine with keeping it v8.

kolyshkin:
For the example of changes needed, see https://github.com/opencontainers/runc/pull/5312.

@adrianreber can you please tag a release? Or should I do it?

snprajwal:
@kolyshkin I've made a v8.3.0 release

Issue #144, opened 2023-08-02
crit: parse data fields in files.img


rst0git:
The Go version of CRIT currently outputs raw data decoded from `files.img`. However, it would be more user-friendly if we map constants used in certain fields to human readable strings and decode the value used for IP address. The following example shows a comparison between the output of the Go and Python versions of CRIT:

* Go version of CRIT
```JSON
        {
            "type": "INETSK",
            "id": 36,
            "isk": {
                "id": 36,
                "ino": 106886,
                "family": 2,
                "type": 1,
                "proto": 6,
                "state": 7,
                "srcPort": 0,
                "dstPort": 0,
                "flags": 2,
                "backlog": 0,
                "srcAddr": [
                    0
                ],
                "dstAddr": [
                    0 
                ],
                "fown": {
                    "uid": 0,
                    "euid": 0,
                    "signum": 0,
                    "pidType": 0,
                    "pid": 0
                },
```

* Python version of CRIT
```JSON
        {
            "type": "INETSK",
            "id": 36,
            "isk": {
                "id": 36,
                "ino": 106886,
                "family": "INET",
                "type": "STREAM",
                "proto": "TCP",
                "state": "CLOSE",
                "src_port": 0,
                "dst_port": 0,
                "flags": "0x2",
                "backlog": 0,
                "src_addr": [
                    "0.0.0.0"
                ],
                "dst_addr": [
                    "0.0.0.0"
                ],
                "fown": {
                    "uid": 0,
                    "euid": 0,
                    "signum": 0,
                    "pid_type": 0,
                    "pid": 0
                },

```

snprajwal:
Hmm, this will require some reworking of the marshaling and unmarshaling logic that we're using currently. We'll probably have to replace `protojson.Marshal` with our own custom logic to marshal it field by field. I'll look into this.

Parthiba-Hazra:
@rst0git if this issue is still on, can I work on this?

rst0git:
@Parthiba-Hazra sure, please feel free to open a pull request!

Parthiba-Hazra:
@rst0git is there any documentations for all types of constants ? -

```
// Constants for families
const (
	AF_INET  = 2
	AF_INET6 = 10
)

// Constants for socket types
const (
	SOCK_STREAM = 1
	SOCK_DGRAM  = 2
)

// Constants for protocols
const (
	IPPROTO_TCP = 6
	IPPROTO_UDP = 17
)

[...]

PR #247, opened 2026-06-06
crit: humanize INETSK fields in files.img JSON output

greenblade29:
## Summary
- When `crit decode --pretty` or `crit show` decodes `files.img`, INET socket (`FdTypes_INETSK`) entries now print human-readable metadata instead of raw protobuf integers.
- Reuses existing helpers from `crit x sk` (`processIP`, `getSkState`, `getAddressFamily`, etc.).
- Adds unit tests for the humanized JSON path (`display_files_test.go`).
Fixes the display gap described in #144 for INET sockets. Example:

| Field | Before | After |
|-------|--------|-------|
| family | `2` | `"INET"` |
| proto | `6` | `"TCP"` |
| state | `7` | `"CLOSE"` |
| src_addr | `[0]` | `["0.0.0.0"]` |
| flags | `2` | `"0x2"` |

## Implementation notes
- `crit.Decode()` calls `applyHumanize()` when `pretty` is true (`crit show` / `crit decode --pretty`).
- `CriuEntry.MarshalJSON()` routes `*fdinfo.FileEntry` through `marshalFileEntryHuman()` when `Humanize` is set.
- Other entry types and non-pretty decode are unchanged (still use `protojson`).
- `crit encode` is unchanged; humanized JSON is output-only and not intended for round-trip encode yet.
## Scope
First slice of #144: **INETSK entries in `files.img` only**. UNIXSK, reg files, and other fd types can follow in separate PRs.

### DIFF
--- a/crit/display_files.go
+++ b/crit/display_files.go
@@ -1,11 +1,19 @@
 package crit
 
 import (
+	"bytes"
+	"encoding/binary"
 	"encoding/json"
 	"fmt"
+	"net/netip"
+	"strconv"
+	"strings"
 
 	"github.com/checkpoint-restore/go-criu/v8/crit/images/fdinfo"
+	"github.com/checkpoint-restore/go-criu/v8/crit/images/memfd"
+	"github.com/checkpoint-restore/go-criu/v8/crit/images/regfile"
 	sk_inet "github.com/checkpoint-restore/go-criu/v8/crit/images/sk-inet"
+	sk_opts "github.com/checkpoint-restore/go-criu/v8/crit/images/sk-opts"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
 )
@@ -16,7 +24,7 @@ var fileEntryProtojson = protojson.MarshalOptions{
 	AllowPartial:  true,
 }
 
-// fileEntryPayloadHumanizer humanizes one FileEntry payload sub-message (reg, isk, …).
+// fileEntryPayloadHumanizer humanizes one FileEntry payload sub-message (reg, isk, ...).
 type fileEntryPayloadHumanizer func(proto.Message) ([]byte, error)
 
 type fileEntryHumanSpec struct {
@@ -28,12 +36,21 @@ type fileEntryHumanSpec struct {
 // fileEntryHumanizers maps fd_types to payload humanizers. Add an entry here when
 // implementing humanized output for a new files.img fd type.
 var fileEntryHumanizers = map[fdinfo.FdTypes]fileEntryHumanSpec{
+	fdinfo.FdTypes_REG: {
+		jsonField: "reg",
+		payload:   regFilePayload,
+		humanize:  humanizeRegPayload,
+	},
 	fdinfo.FdTypes_INETSK: {
 		jsonField: "isk",
-		payload:   func(e *fdinfo.FileEntry) proto.Message { return e.GetIsk() },
+		payload:   inetSkPayload,
 		humanize:  humanizeInetSkPayload,
 	},
-	// fdinfo.FdTypes_REG: { jsonField: "reg", payload: ..., humanize: humanizeRegPayload },
+	fdinfo.FdTypes_MEMFD: {
+		jsonField: "memfd",
+		payload:   memfdFilePayload,
+		humanize:  humanizeMemfdPayload,
+	},
 }
 
 var (
@@ -69,7 +86,51 @@ var (
 		9:  "LAST_ACK",
 		10: "LISTEN",
 	}
+	rfileFlagHumanMap = []flagHumanEntry{
+		{name: "O_WRONLY", value: 0o00000001},
+		{name: "O_RDWR", value: 0o00000002},
+		{name: "O_CREAT", value: 0o00000100},
+		{name: "O_EXCL", value: 0o00000200},
+		{name: "O_NOCTTY", value: 0o00000400},
+		{name: "O_TRUNC", value: 0o00001000},
+		{name: "O_APPEND", value: 0o00002000},
+		{name: "O_NONBLOCK", value: 0o00004000},
+		{name: "O_DSYNC", value: 0o00010000},
+		{name: "FASYNC", value: 0o00020000},
+		{name: "O_DIRECT", value: 0o00040000},
+		{name: "O_LARGEFILE", value: 0o00100000},
+		{name: "O_DIRECTORY", value: 0o00200000},
+		{name: "O_NOFOLLOW", value: 0o00400000},
+		{name: "O_NOATIME", value: 0o01000000},
+		{name: "O_CLOEXEC", value: 0o02000000},
+	}
 )
+
+type flagHumanEntry struct {
+	name  string
+	value uint64
+}
+
+func regFilePayload(entry *fdinfo.FileEntry) proto.Message {
+	if payload := entry.GetReg(); payload != nil {
+		return payload
+	}
+	return nil
+}
+
+func inetSkPayload(entry *fdinfo.FileEntry) proto.Message {
+	if payload := entry.GetIsk(); payload != nil {
+		return payload
+	}
+	return nil
+}
+
+func memfdFilePayload(entry *fdinfo.FileEntry) proto.Message {
+	if payload := entry.GetMemfd(); payload != nil {
+		return payload
+	}
+	return nil
+}
 
 // fileEntryJSON mirrors file_entry field order from fdinfo.proto.
 type fileEntryJSON struct {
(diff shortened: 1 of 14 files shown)


crit/image.go
```diff
@@ -34,19 +36,37 @@ func (c *CriuEntry) MarshalJSON() ([]byte, error) {
...
-		extraString := fmt.Sprint(`"extra":"`, c.Extra, `"}`)
-		data[len(data)-1] = byte(',')
-		data = append(data, []byte(extraString)...)
+		return appendExtraJSON(data, c.Extra)
 	}
 	return data, nil
 }
 
+func appendExtraJSON(data []byte, extra string) ([]byte, error) {
+	extraString := fmt.Sprint(`"extra":"`, extra, `"`)
```
adrianreber: The original line had a `}` at the end. This seems to be gone here.

crit/display_files.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+type inetSkJSON struct {
+	ID      uint32   `json:"id,omitempty"`
+	Ino     uint32   `json:"ino,omitempty"`
+	Family  string   `json:"family,omitempty"`
+	Type    string   `json:"type,omitempty"`
+	Proto   string   `json:"proto,omitempty"`
+	State   string   `json:"state,omitempty"`
+	SrcPort uint32   `json:"src_port,omitempty"`
```
adrianreber: A couple of things here (SrcPort, DstPort, Backlog, ID, and Ino) can be set to `0`. Omitting them seems wrong.

crit/display_files_test.go
```diff
@@ -0,0 +1,101 @@
...
+		Id:   proto.Uint32(36),
+		Type: fdinfo.FdTypes_INETSK.Enum(),
+		Isk: &sk_inet.InetSkEntry{
+			Id:      proto.Uint32(36),
+			Ino:     proto.Uint32(106886),
+			Family:  proto.Uint32(2), // AF_INET
+			Type:    proto.Uint32(1), // SOCK_STREAM
+			Proto:   proto.Uint32(6), // TCP
+			State:   proto.Uint32(7), // TCP_CLOSE
+			SrcPort: proto.Uint32(0),
```
adrianreber, line 29: You are setting SrcPort to "0" in the test. It should not appear in the json because of omitempty, right?

crit/display_files.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/checkpoint-restore/go-criu/v8/crit/images/fdinfo"
+	sk_inet "github.com/checkpoint-restore/go-criu/v8/crit/images/sk-inet"
+	"google.golang.org/protobuf/proto"
+)
+
+// fileEntryJSONis a human-readable view of files.img: to be used with criu show/decode
```
adrianreber: space before "is"

adrianreber:
Also add a bats based test for this functionality.

adrianreber:
I would prefer if you do not create a link in the commit message to the issue (do not put a `#` in front of the issue number or something). This will pop up all over the place in the future. Just mention it in the PR description.

Please mention which AI tool you have used to create this in the commit message like described in https://github.com/checkpoint-restore/go-criu?tab=contributing-ov-file

crit/display_files_test.go
greenblade29: I have fixed it in the JSON struct instead, you're right: it should not have been `omitempty` in the first place.

crit/image.go
greenblade29: You're right, I missed it.

rst0git:
@greenblade29 It might be helpful to copy/paste the relevant output of `crit show` or `crit decode` in the commit message as an example.

crit/display_files.go
```diff
@@ -0,0 +1,127 @@
...
+type inetSkJSON struct {
+	ID      uint32   `json:"id"`
+	Ino     uint32   `json:"ino"`
+	Family  string   `json:"family,omitempty"`
+	Type    string   `json:"type,omitempty"`
+	Proto   string   `json:"proto,omitempty"`
+	State   string   `json:"state,omitempty"`
+	SrcPort uint32   `json:"src_port"`
+	DstPort uint32   `json:"dst_port"`
+	Flags   string   `json:"flags,omitempty"`
```
rst0git: It looks like the `flags` field is `uint32`:
https://github.com/checkpoint-restore/criu/blob/df9f9a6cd59482eee9c97e9176c058a4dec04fbf/images/sk-inet.proto#L44

crit/display_files.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+	case fdinfo.FdTypes_PACKETSK:
+		if psk := entry.GetPsk(); psk != nil {
+			out.Psk = psk
+		}
+	case fdinfo.FdTypes_NETLINKSK:
+		if nlsk := entry.GetNlsk(); nlsk != nil {
+			out.Nlsk = nlsk
+		}
+	case fdinfo.FdTypes_EXT:
```
rst0git: `"type": "MEMFD"` and a few others are missing: https://github.com/checkpoint-restore/criu/blob/df9f9a6cd59482eee9c97e9176c058a4dec04fbf/images/fdinfo.proto#L44

rst0git:
@greenblade29 Our goal is for the Go version of CRIT to produce an equivalent output to the [Python version](
https://github.com/checkpoint-restore/criu/tree/criu-dev/crit). For example, you can use the ZDTM tests for CRIU to create images:
```
$ cd criu/test
$ sudo ./zdtm.py run -t zdtm/static/socket-tcpbuf -k always --ignore-taint
```
Then, compare the output with both CRIT versions:
```
$ sudo crit decode --pretty -i dump/zdtm/static/socket-tcpbuf/*/*/files.img > /tmp/test-1.json
```
```diff
--- a/tmp/files-1.json
+++ b/tmp/files-2.json
@@ -2,11 +2,11 @@
     "magic": "FILES",
     "entries": [
         {
-            "type": "REG",
             "id": 1,
+            "type": "REG",
             "reg": {
                 "id": 1,
-                "flags": "",
+                "flags": 0,
                 "pos": 0,
                 "fown": {
                     "uid": 0,
@@ -28,11 +28,11 @@
             }
         },
         {
-            "type": "REG",
             "id": 2,
+            "type": "REG",
             "reg": {
                 "id": 2,
-                "flags": "",
+                "flags": 0,
                 "pos": 0,
                 "fown": {
                     "uid": 0,
@@ -54,28 +54,15 @@
             }
         },
         {
-            "type": "MEMFD",
             "id": 3,
-            "memfd": {
-                "id": 3,
-                "flags": "",
-                "pos": 0,
-                "fown": {
-                    "uid": 0,
-                    "euid": 0,
-                    "signum": 0,
-                    "pid_type": 0,
-                    "pid": 0
-                },
-                "inode_id": 1
-            }
+            "type": "MEMFD"
         },
         {
-            "type": "REG",
             "id": 4,
+            "type": "REG",
             "reg": {
                 "id": 4,
-                "flags": "",
+                "flags": 0,
                 "pos": 0,
                 "fown": {
                     "uid": 0,

```

crit/display_files.go
greenblade29: I have added all of them as cases now.

crit/display_files.go
greenblade29: I modified the implementation a bit to match Python output. Now we are using `json.RawMessage` everywhere: reason being, humanization is now a patch on top of protojson. We won't be doing a full re-decode of every nested message, instead we use the output from protojson and replace the fields that CRIT humanizes.

adrianreber:
Thanks for adding a bats based test. I kind of forgot that this project does not have any bats based testing. I was thinking about something else. Thanks for adding it.

test/crit/Makefile
```diff
@@ -25,14 +31,20 @@ define gen_mmapper_imgs
...
-test-imgs: ../loop/loop ../mmapper/mmapper
+test-imgs: ../loop/loop ../mmapper/mmapper inetsk_sleeper/inetsk_sleeper
 	$(eval LOOP_PID := $(shell ../loop/loop))
 	$(eval LOOP_DIR := $@/loop)
 	mkdir -p ${LOOP_DIR}
 	cp /proc/${LOOP_PID}/environ ${LOOP_DIR}
 	cp /proc/${LOOP_PID}/cmdline ${LOOP_DIR}
 	$(CRIU) dump -v4 -o dump.log -D ${LOOP_DIR} -t $(LOOP_PID)
 
+	$(eval INETSK_PID := $(shell ./inetsk_sleeper/inetsk_sleeper))
```
adrianreber, line 42: Your test process hangs here and breaks CI.

test/crit/Makefile
```diff
@@ -25,14 +31,20 @@ define gen_mmapper_imgs
...
+	$(eval INETSK_DIR := $@/inetsk)
+	mkdir -p ${INETSK_DIR}
+	$(CRIU) dump -v4 -o dump.log -D ${INETSK_DIR} -t $(INETSK_PID)
+	pkill -9 inetsk_sleeper || true
```
adrianreber: You already have the PID. Use that and do not kill all processes with the same name.

rst0git (approved):
LGTM

@greenblade29 I've updated the pull request to resolve the outstanding comments.

Issue #254, opened 2026-07-07
Add ImageIoMode to go-criu


dfeigin-nv:
When https://github.com/checkpoint-restore/criu/pull/3066 is merged, --image-io-mode is not automatically pipelined into go-criu. would it be possible to add it?

adrianreber:
Sure, just open a PR.

dfeigin-nv:
Done. Will it be possible to update go-criu version afterwards?

adrianreber:
A release is always possible. No problem. Especially if it doesn't break API.

PR #255, opened 2026-07-08
rpc: add image_io_mode option (#254)

dfeigin-nv:
Adds the `--image-io-mode` RPC option to go-criu (#254). CRIU [PR #3066](https://github.com/checkpoint-restore/criu/pull/3066) adds the `criu_image_io_mode` enum and `image_io_mode = 72 [default = IMAGE_IO_WRITEBACK]` to `criu_opts`. go-criu generates its bindings from CRIU's `rpc.proto`, so the field has to be mirrored here for callers to set `opts.ImageIoMode`.

- `rpc/rpc.proto` — add the enum and field. Field 72 matches #3066.
- `rpc/rpc.pb.go` — regenerated.

### DIFF
--- a/rpc/rpc.proto
+++ b/rpc/rpc.proto
@@ -60,6 +60,11 @@ enum criu_pre_dump_mode {
 	VM_READ =	2;
 };
 
+enum criu_image_io_mode {
+	IMAGE_IO_WRITEBACK	= 0;
+	IMAGE_IO_DIRECT	= 1;
+};
+
 message criu_opts {
 	required int32			images_dir_fd	= 1 [default = -1];
 	optional string			images_dir	= 68; /* used only if images_dir_fd == -1 */
@@ -145,6 +150,7 @@ message criu_opts {
 	optional bool			leave_stopped		= 69;
 	optional bool			display_stats		= 70;
 	optional bool			log_to_stderr		= 71;
+	optional criu_image_io_mode	image_io_mode		= 72 [default = IMAGE_IO_WRITEBACK];
 /*	optional bool			check_mounts		= 128;	*/
 }
 
--- a/rpc/rpc.pb.go
+++ b/rpc/rpc.pb.go
@@ -133,6 +133,39 @@ func (x CriuPreDumpMode) String() string {
 	return strconv.Itoa(int(x))
 }
 
+type CriuImageIoMode int32
+
+const (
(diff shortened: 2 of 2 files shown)


adrianreber:
So, just an update of the protobuf files? Nothing else? No functions needed for your use case? Just curious. Not a blocker.

dfeigin-nv:
Yes I want to be able to set opts.ImageIoMode when calling Checkpoint/Restore from Snapshot. This specific PR is just to pipe those options into criu since https://github.com/checkpoint-restore/criu/pull/3066 is merged. Unless there is something I missed and there is an additional change required?

adrianreber:
No, all good. Sometimes we have additional code in go-criu for some functionality. If it is not necessary then we are good.

adrianreber:
How did you do this change? I haven't done it in a long time and I am not 100% sure any more how we used to handle this.

I am confused about this change:

```diff
// Code generated by protoc-gen-go-lite. DO NOT EDIT.
// protoc-gen-go-lite version: v0.14.0
-// source: rpc.proto
+// source: rpc/rpc.proto

package rpc
```

rpc/rpc.proto
```diff
@@ -60,6 +60,11 @@ enum criu_pre_dump_mode {
 	VM_READ =	2;
 };
 
+enum criu_image_io_mode {
+	IMAGE_IO_WRITEBACK = 0;
+	IMAGE_IO_DIRECT    = 1;
```
rst0git: We usually sync the `rpc.proto` file with the master branch of CRIU after CRIU release
(e.g., `rm rpc/rpc.proto && make rpc/rpc.proto`).

Would it be possible to copy the file from the criu-dev branch to keep it consistent? There are some white-space changes here.

dfeigin-nv: Certainly

dfeigin-nv:
@adrianreber Radostin pointed out how to properly do this chagne  (copy the proto from criu-dev and regenerate it with make).

rst0git (approved):
LGTM

adrianreber:
> @adrianreber Radostin pointed out how to properly do this chagne (copy the proto from criu-dev and regenerate it with make).

Perfect. The Makefile only supports from the CRIU release and not development branch. But this looks good now. Thanks everyone.

PR #259, opened 2026-07-16
Add support for compressed CRIU images

rst0git:
In CRIU's development branch was recently added built-in support for compressing memory pages during checkpoint creation and decompressing them during restore [^1][^2][^3]. As this functionality changes the memory image format, the Python implementation of CRIT was also extended with commands for compressing and decompressing checkpoint images.

This pull request updates go-criu's RPC and CRIU protobuf definitions to match the criu-dev branch, adds support for reading compressed memory images, and extends the Go implementation of CRIT with equivalent `compress` and `decompress` commands.

This functionality in this pull request is also needed by `checkpointctl` to inspect and analyze checkpoints with compressed memory images.

[^1]: [CRIU pull request "Add support for memory pages compression"](https://github.com/checkpoint-restore/criu/pull/2895)
[^2]: [Towards On-the-Fly Snapshot Memory Compression for Low-Latency Elastic Inference Serving Systems (EuroMLSys '26)](https://doi.org/10.1145/3805621.3807612)
[^3]: [Optimizing Checkpoints with Built-in Memory Page Compression (LPC '25)](https://lpc.events/event/19/contributions/2236/)


rst0git:
cc: @snprajwal @behouba @dfeigin-nv @adrianreber 

Would you be able to review the changes in this PR when you have time?

adrianreber:
I will look at it tomorrow.

crit/mempages.go
```diff
@@ -226,38 +152,32 @@ func (mr *MemoryReader) SearchPattern(pattern string, escapeRegExpCharacters boo
...
-				readSize = int(endAddr - startAddr - offset)
+				readSize = endAddr - startAddr - offset
 			}
 
-			buff := make([]byte, readSize)
-			if _, err := f.ReadAt(buff, int64(initialOffset+offset)); err != nil {
-				if err == io.EOF {
-					break
-				}
+			memory, err := mr.readMemRange(session, startAddr+offset, startAddr+offset+readSize)
```
dfeigin-nv: It doesn't have to do with your PR specifically, but I think there may be a bug here. This searches each non-overlapping chunk independently, so a match crossing a chunk boundary is missed. I may be mistaken but it looks like it.

.github/workflows/codecov.yml
```diff
@@ -16,9 +16,9 @@ jobs:
 
     - name: Build CRIU criu-dev
       run: |
-        sudo apt-get update
-        sudo apt-get install -y libprotobuf-dev libprotobuf-c-dev protobuf-c-compiler protobuf-compiler python3-protobuf libnl-3-dev libnet-dev libcap-dev uuid-dev curl unzip bats
         git clone --depth=1 --single-branch -b criu-dev https://github.com/checkpoint-restore/criu.git
+        sudo ./criu/contrib/dependencies/apt-packages.sh
+        sudo apt-get install -y curl unzip bats
```
adrianreber: You could use the apt-install script from CRIU here. Not sure it is really super helpful, but it might make it more robust against temporary repository failures.

crit/memory_layout.go
```diff
@@ -0,0 +1,573 @@
...
+	"slices"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/checkpoint-restore/go-criu/v8/crit/images/pagemap"
+)
+
+const (
+	peParent         uint32 = 1 << 0
```
adrianreber, line 17: There are a couple of constants here and in transform_linux.go which are almost the same. It is the same Go package so maybe try to reuse constants and functions here if possible. (`effectivePagemapFlags()`, `pagemapPageCount()`, `hasPagemapCompressionMetadata()`, `checkedAlignUp()`).

crit/memory_layout.go
```diff
@@ -0,0 +1,573 @@
...
+					regionBytes,
+					info.RegionSizeBytes,
+				),
+			)
+		}
+	}
+	return nil
+}
+
+func indexMemoryLayer(
```
adrianreber, line 164: Is this the same as `transformValidatePagemaps()`?

crit/checkpoint_compression_impl.go
```diff
@@ -0,0 +1,1491 @@
...
+	}
+	return merged
+}
+
+func transformAddressInRanges(address uint64, ranges []transformRange) bool {
+	index := sort.Search(len(ranges), func(i int) bool { return ranges[i].start > address }) - 1
+	if index >= 0 && address < ranges[index].end {
+		return true
+	}
+	if address>>32 == math.MaxUint32 {
```
adrianreber: What is happening here? Overall it feels really hard to understand with almost zero comments what is going on in these files.

crit/checkpoint_compression_impl.go
```diff
@@ -0,0 +1,1491 @@
...
+			break
+		}
+		replaced = append(replaced, item)
+	}
+	if commitErr == nil {
+		commitErr = transformSyncDirectories(staged)
+		if commitErr == nil {
+			// A cancellation received through the final rename or directory
+			// sync still belongs to the commit window and must roll back.
+			commitErr = ctx.Err()
```
adrianreber: At this point all renames have succeeded and all directories have been fsynced, right? Shouldn't the transform be considered fully committed here, rather than rolling back a durable result because of a cancellation signal that arrived after  the work was already persisted?

I am really a bit confused about this large file. Some additional documentation would be nice.

crit/mempages.go
rst0git: This is a good point! Indeed, it looks like the search functionality would miss patterns that cross a chunk boundary. I've opened another pull request to fix this: https://github.com/checkpoint-restore/go-criu/pull/260

.github/workflows/codecov.yml
rst0git: Thanks! I've updated the patch to use `apt-install`.

crit/memory_layout.go
rst0git: Thanks! I've updated the patch to remove duplicate definitions.

crit/memory_layout.go
rst0git: Yes, most of the validation was duplicated. I've updated the patch to reuse `indexMemoryLayer()` and the shared inventory compatibility validation.

crit/checkpoint_compression_impl.go
rst0git: Thanks! I've updated the code to add comments.

The Go implementation here is equivalent to the way the Python version of CRIT is handling [this](https://github.com/checkpoint-restore/criu/blob/b7bdc14736b1e8598f2ffbb371f42318835a111f/crit/crit/__main__.py#L589).

My understanding is that on x86, CRIU's [encode_pointer()](https://github.com/checkpoint-restore/criu/blob/b7bdc14736b1e8598f2ffbb371f42318835a111f/criu/arch/x86/include/asm/types.h#L35) casts pointers through a signed `long`, and pagemap writers store that encoded value. In contrast, CRIU's VMA parser stores addresses parsed from `/proc/<pid>/smaps` as [unsigned values](https://github.com/checkpoint-restore/criu/blob/b7bdc14736b1e8598f2ffbb371f42318835a111f/criu/proc_parse.c#L820). As a result, native 32-bit addresses may be sign-extended in pagemap images but zero-extended in `mm.img`.

crit/checkpoint_compression_impl.go
rst0git: Thanks! I removed the post-fsync cancellation check, so the final successful directory sync is now the commit point. I also comments about the transaction phases.

adrianreber (approved):
Some higher coverage by tests would be nice. It seems like mostly error paths are missing.

Besides that looks good to go.

We need to wait with a release until CRIU has a release because changes from the `criu-dev` branch to the protobuf files are not necessarily stable. Once it is part of a CRIU release it will not change.

adrianreber:
Just to clarify, from my point of view we can merge it now.

avagin:
The CRIU part is still in criu-dev, so there is no guarantee that UAPI will not be changed. Ideally, all changes in go-criu should be made only after merging CRIU changes to the master branch...

rst0git:
Thank you Andrei! One of the main use cases for the compression functionality has been to improve the cold-start times of inference workloads. Many of the open-source projects in this space evolve rapidly, with new models being released almost every month.

I spent more than a year working on the compression functionality as part of my PhD research, and did my best to design an API that is minimal and would allow us to support future developments.

Since we don't have a regular release cycle for CRIU, it makes it difficult for other projects and users to plan ahead and know when new features will become available.

avagin:
@rst0git I think we already discussed this in personal discussions. I am ready to release this feature as soon as it is ready. We are quite flexible with our release process. We are not tied to any specific timeline and can cut a new release even today, if needed.

Any projects using the criu-dev branch must accept the risk that things may change and backward compatibility is not guaranteed. The only source of truth is the master branch.

PR #260, opened 2026-07-17
crit: find pattern matches across memory chunk boundaries

rst0git:
`MemoryReader.SearchPattern` currently searches each memory chunk independently, which means patterns that span chunk boundaries are missed. This pull request aims to fix this by searching exact literals with overlapping chunks, while evaluating general regular expressions through a bounded streaming reader over the entire pagemap entry.

### DIFF
--- a/crit/mempages.go
+++ b/crit/mempages.go
@@ -539,7 +539,15 @@ func (mr *MemoryReader) SearchPattern(pattern string, escapeRegExpCharacters boo
 		return nil, err
 	}
 	literalPattern, literalOnly := regexPattern.LiteralPrefix()
+	needsStreaming := !literalOnly || len(literalPattern) == 0
 	var patternTable []int
+	var streamPatterns *streamingRegexps
+	if needsStreaming {
+		streamPatterns, err = compileStreamingRegexps(pattern)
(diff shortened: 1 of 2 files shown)


rst0git:
cc: @behouba

behouba (approved):
LGTM

crit/mempages.go
```diff
@@ -204,6 +205,288 @@ type PatternMatch struct {
...
+		readEnd += min(overlap, entrySize-readEnd)
+		readSize := readEnd - readStart
+		if readSize > uint64(math.MaxInt) {
+			return nil, fmt.Errorf("memory search window is too large: %d bytes", readSize)
+		}
+		buff := make([]byte, int(readSize))
+		if _, err := f.ReadAt(buff, int64(initialOffset+readStart)); err != nil {
+			if err == io.EOF {
+				break
+			}
```
behouba:
```suggestion
			if errors.Is(err, io.EOF) {
				err = io.ErrUnexpectedEOF
			}
```

Can we make this one return an error too?  The streaming path `searchPatternStream` handles the same situation by returning io.ErrUnexpectedEOF.

PR #269, opened 2026-08-27
dependabot: group dependency updates

snprajwal:
Dependabot can group updates for a package ecosystem into a single pull request, making it easier to review and merge dependency updates. Add the `gomod` and `github-actions` groups for the respective package ecosystems.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,7 +4,15 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    groups:
+      gomod:
+        patterns:
+          - "*"
   - package-ecosystem: github-actions
     directory: /
     schedule:
       interval: weekly
+    groups:
+      github-actions:
+        patterns:
+          - "*"


snprajwal:
To note, this was previously being used for just the `gomod` package ecosystem, and was removed in ea940e0c6325177bb92478c61372c15c68a305b5. The commit message says it was done for simpler PR management, but I'm not sure how it's simpler to have many different PRs that need to be individually approved and merged. If there are other concerns about this feature, please let me know.

adrianreber:
I don't like group updates. They are even harder to read then single vendor updates. That is just me. If more people prefer group updates let's do that. But there are not that many updates here.

snprajwal:
Thanks for clarifying. I agree that it used to be much messier, it looks like a recent update to Dependabot now makes it use a clearer table-based format for multiple updates (for example, in [this PR](https://github.com/checkpoint-restore/checkpointctl/pull/226)).

From [the PR history](https://github.com/checkpoint-restore/go-criu/pulls?q=is%3Apr+is%3Aclosed+author%3Aapp%2Fdependabot) on this repo, there are 5-6 Dependabot PRs per month. Grouping would hopefully reduce that in cases where multiple packages are updated in the same week. I don't have a strong opinion either, happy to keep the existing individual PR model if that works better.

